PR #36, opened 2026-03-23
Planning pushpullhook

Jaraxxus-Me:
I made a few changes to make bilevel planning kinda work in pushpullhook2d-o0. The sim still has a resettable issue, but it now has a reasonable mismatch that planning could sometimes work with stochastic physics. See videos here.

https://github.com/user-attachments/assets/7814a08c-2dec-4bdb-b75f-b01362844131


yixuanhuang98 (approved):
LGTM! Could you also add an example for DynPushPullHook2D-o5 because we ran other baselines on the o5 variant? thanks!

PR #53, opened 2026-04-07
Added in-context examples for tasks

vaibhavsaxena11:


yixuanhuang98 (commented):
Should we put in-context examples in the kinder-baselines instead of kindergarden?

vaibhavsaxena11:
> Should we put in-context examples in the kinder-baselines instead of kindergarden?

Yes, actually that would be better. Please take a look at my other PR [here](https://github.com/Princeton-Robot-Planning-and-Learning/kinder-baselines/pull/24) for LLMPlanner support in kinder-baselines. I'll create a subsequent PR to port changes from here to another PR in kinder-baselines.

PR #54, opened 2026-04-08
Ensure stable reset

yixuanhuang98:
Added several env.step() during reset() function to ensure that the objects will be stable on the ground or on the table. This can ensure that the downstream controllers get a stable object pose.


vaibhavsaxena11 (approved):
LGTM

PR #88, opened 2026-04-30
Fix Packing3D triangle drift during state restoration

siegelz:
## Summary

This PR fixes a state restoration bug in Packing3D where triangular parts drift each time an object-centric state is restored with set_state().

## Bug

For triangular Packing3D parts, Packing3DObjectCentricState.get_object_pose() returns a semantic pose whose position is shifted by the triangle centroid offset. That is useful for callers that want the pose of the triangular prism center, but it is not the raw PyBullet body pose stored in the state features.

ObjectCentricPacking3DEnv._set_object_states() was using obs.get_object_pose(name) to restore part bodies. For triangles, that wrote the centroid-shifted semantic pose back into PyBullet as if it were the raw pose. As a result, repeated state round trips accumulated the centroid offset:

```python
obs, _ = env.reset(seed=123)
for _ in range(3):
    env.set_state(obs)
    obs = env.get_state()
```

After each iteration, a triangular part's stored pose_x / pose_y changed even though no action was applied.

## Fix

The restore path now reads the raw pose features directly from the state:

- pose_x, pose_y, pose_z
- pose_qx, pose_qy, pose_qz, pose_qw

This keeps the public get_object_pose() semantics unchanged while making set_state(get_state()) idempotent for triangular parts.

## Test

Adds test_triangle_part_set_state_does_not_drift, which repeatedly restores a one-part Packing3D state and verifies that the triangular part's stored pose remains unchanged.

Validated with:

```bash
pytest tests/envs/kinematic3d/test_packing3d.py::test_triangle_part_set_state_does_not_drift
pytest tests/envs/kinematic3d/test_packing3d.py
```


tomsilver (approved):
Thanks so much for the contribution! :)

PR #95, opened 2026-05-17
fixed issue81, using shapely.Polygon to check goals in Packing3D

Jaraxxus-Me:
Addressed this issue: https://github.com/Princeton-Robot-Planning-and-Learning/kindergarden/issues/89

We are now using shapely.polygon.contain to check the goal condition, instead of `inside`, I added a unit test function where `is_inside` returns false, but the goal is actually achieved.


tomsilver (approved):
LGTM!

PR #97, opened 2026-05-21
Fix held hook collision type when resetting DynScoopPour2D state

siegelz:
## Summary

Fix a `DynScoopPour2D` mismatch between live grasped hooks and held hooks reconstructed from an object-centric state.

## Issue

In the live grasp path, [`on_gripper_grasp()` copies the original hook shapes into a new kinematic body](https://github.com/Princeton-Robot-Planning-and-Learning/kindergarden/blob/main/src/kinder/envs/dynamic2d/utils.py#L904-L908). The code explicitly leaves the `ROBOT_COLLISION_TYPE` reassignment commented out:

```python
copied_shape = shape.copy()
copied_shape.body = kinematic_body
# copied_shape.collision_type = ROBOT_COLLISION_TYPE
```

As a result, a live-held hook is a `pymunk.Body.KINEMATIC` body whose shapes retain their original `DYNAMIC_COLLISION_TYPE`.

The reset/reconstruction path in `DynScoopPour2D` used different semantics: when `hook.held == True`, it rebuilt the hook as kinematic but assigned its shapes `ROBOT_COLLISION_TYPE`. That made reset/replay/particle rollout differ from live rollout for the same logical held-hook state.

## Manifestation

When a held hook overlaps static geometry such as the middle wall, the reconstructed hook could trigger robot/static collision rollback behavior that the live-held hook would not. This can cause replay or particle rollout divergence, including the hook jumping, disappearing, or otherwise becoming inconsistent after reset.

## Fix

Make the reconstruction path match live grasp behavior. The held hook body remains kinematic, but [its shapes now use `DYNAMIC_COLLISION_TYPE`](https://github.com/Princeton-Robot-Planning-and-Learning/kindergarden/pull/97/files#diff-3bd00b2b2e0ed88510207d55150487d5036e55c560b9eb93dbd0c6c085e5b10eR453-R457).

This does **not** make the held hook a dynamic body. It only makes the shape collision category consistent:

- live held hook: `KINEMATIC` body, `DYNAMIC_COLLISION_TYPE` shapes
- reset held hook: `KINEMATIC` body, `DYNAMIC_COLLISION_TYPE` shapes

## Scope

This is intentionally narrow. It does not change generic motion planning or global held-object semantics; it only fixes `DynScoopPour2D` state reconstruction consistency.

## Validation

- Ran downstream focused program tests with this submodule change: `pixi run pytest tests/unit/test_programs.py -q`
- Ran a downstream seeded `DynScoopPour2D` rollout to verify reset/replay no longer crashes inference


Jaraxxus-Me (approved):
LGTM!
(But this will make the task significantly harder lol, I was not able to collect one demo with this collision setting)

siegelz:
@Jaraxxus-Me How did this change the behavior of collecting the demo? The hook is still supposed to be able to pass through the wall, did this not happen for you?

PR #98, opened 2026-05-21
Fix Packing3D triangle available-parts containment check

siegelz:
## Summary

Fix `Packing3DObjectCentricState.available_parts` so triangle parts use the same triangle-aware rack containment check as `goal_reached()`.

## Issue

`Packing3D` already has `is_part_inside_rack()`, which checks a part's true footprint against the rack footprint. This handles triangular parts correctly.

However, `available_parts` was still using the older `is_inside(...)` check with half extents. For triangular parts, that uses a bounding-box-style containment test rather than the true triangle footprint. This could mark a triangle as still available even when its actual footprint was inside the rack.

## Fix

Update `available_parts` to call `self.is_part_inside_rack(obj.name)` instead of duplicating the old half-extents containment check.

The existing `test_packing3d_goal` now also asserts that the placed triangle is not listed in `available_parts`.

## Validation

- `pixi run pytest lib/kindergarden/tests/envs/kinematic3d/test_packing3d.py::test_packing3d_goal -q`


tomsilver (approved):
thanks again for these contributions!

siegelz:
@tomsilver I don't believe I have write access to this repo, do you need to merge it?

PR #103, opened 2026-05-26
Add point cloud generation for Dynamic3D MuJoCo environments

yixuanhuang98:
## Summary

Adds point cloud generation for Dynamic3D MuJoCo environments, enabling dense RGB+XYZ scene representations from any set of named cameras via depth rendering. Also provides tooling to replay demonstrations and export both RGB images and point clouds to HDF5, plus an Open3D visualization script for inspecting the saved data.

### `src/kinder/envs/dynamic3d/point_cloud.py`

Core module with the following public API:

| Symbol | Description |
|--------|-------------|
| `PointCloud` | Dataclass holding `xyz (N,3) float32`, `rgb (N,3) uint8`, `camera_indices (N,)` and `camera_names`. Supports `filter_by_camera()` and `to_dict()` for serialisation. |
| `get_camera_intrinsics(model, cam_id, w, h)` | Builds the 3×3 intrinsic matrix K from `cam_fovy` (square-pixel MuJoCo cameras). |
| `get_camera_extrinsics(data, cam_id)` | Returns the camera-to-world rotation `R` and translation `t` from `cam_xmat` / `cam_xpos`. |
| `depth_buffer_to_metric(depth_raw, model)` | Converts raw MuJoCo OpenGL depth buffer values to metric distances: `z = znear·zfar / (zfar − d·(zfar − znear))`. |
| `rgbd_to_point_cloud(rgb, depth, cam_id, model, data)` | Back-projects one RGBD image pair into world-frame xyz+rgb, discarding pixels outside `[min_depth, max_depth]`. |
| `generate_scene_point_cloud(sim, camera_names, ...)` | Renders all (or selected) cameras through `MjRenderContextOffscreen` and merges their back-projections into a single `PointCloud`. |
| `get_sim_from_env(env)` | Extracts the `MjSim` from any wrapped kinder gymnasium env. |

### `tests/envs/dynamic3d/test_point_cloud.py` *(replaces `scripts/visualize_point_clouds.py`)*

27 unit tests covering the full public API:

| Area | Tests |
|------|-------|
| `get_sim_from_env` | returns sim; raises before reset |
| `get_camera_intrinsics` | shape, pinhole structure, square pixels |
| `get_camera_extrinsics` | shapes, orthogonal rotation matrix |
| `depth_buffer_to_metric` | zero→near plane, positive values, shape preserved |
| `rgbd_to_point_cloud` | shapes, dtypes, empty when no valid pixels |
| `generate_scene_point_cloud` | type, non-empty, array shapes/dtypes, RGB range, camera index validity, single-camera restriction, unknown camera error |
| `PointCloud` methods | `filter_by_camera`, unknown camera error, `to_dict`, `len` |

Two optional pytest flags (registered in `tests/conftest.py`) allow interactive inspection without a separate script:

```bash
# Open matplotlib RGB renders, depth maps, and 3-D scatter plot
pytest tests/envs/dynamic3d/test_point_cloud.py --visualize -s

# Save point cloud as .npz for offline viewing
pytest tests/envs/dynamic3d/test_point_cloud.py \
    --save-point-cloud /tmp/scene_pc.npz -s
```

### `tests/conftest.py`

Added two project-wide pytest options:
- `--visualize` — enable matplotlib visualization tests
- `--save-point-cloud <path>` — save generated point cloud to a `.npz` file

### `scripts/visualize_saved_point_cloud.py`

Standalone script to visualise `.npz` point cloud files saved by `--save-point-cloud`:

```bash
python scripts/visualize_saved_point_cloud.py /tmp/scene_pc.npz
python scripts/visualize_saved_point_cloud.py /tmp/scene_pc.npz --by-camera
python scripts/visualize_saved_point_cloud.py /tmp/scene_pc.npz --max-pts 20000
python scripts/visualize_saved_point_cloud.py /tmp/scene_pc.npz \
    --save-figs /tmp/pc --no-show
```

### `scripts/demos_to_hdf5_with_pointclouds.py`

Replays demonstration pickle files and writes per-step RGB images + point clouds to HDF5. Extends the same concept as `kinder-baselines/kinder-models/scripts/demos_to_hdf5.py` with per-camera point clouds alongside RGB frames.

**HDF5 structure:**
```
data/
  attrs: env_id, total_episodes, total_frames, cameras
  demo_0/
    attrs: num_frames, seed
    obs/
      state           (N, obs_dim)    float32
      {cam}_rgb       (N, H, W, 3)   uint8
      {cam}_pc_xyz    (N, max_pts, 3) float32   (NaN-padded if fewer pts)
      {cam}_pc_rgb    (N, max_pts, 3) uint8     (zero-padded if fewer pts)
    actions           (N, act_dim)   float32
  demo_1/
    ...
```

```bash
# Convert all demos in a directory
python scripts/demos_to_hdf5_with_pointclouds.py \
    --demo-dir demos/SweepIntoDrawer3D-o5 \
    --output /tmp/sweep_pcs.hdf5

# Specific cameras, fixed point cloud size, smaller renders
python scripts/demos_to_hdf5_with_pointclouds.py \
    --demo-dir demos/SweepIntoDrawer3D-o5 \
    --output /tmp/sweep_pcs.hdf5 \
    --cameras agentview_1 robot_wrist \
    --render-width 84 --render-height 84 \
    --max-pts 8192
```

### `scripts/visualize_hdf5_open3d.py`

Open3D-based viewer for interactive point cloud inspection. Opens two windows simultaneously: Open3D (point cloud with orbit/pan/zoom) and matplotlib (RGB images), kept in sync step-by-step.

```bash
python scripts/visualize_hdf5_open3d.py data.hdf5
python scripts/visualize_hdf5_open3d.py data.hdf5 --animate --fps 5
python scripts/visualize_hdf5_open3d.py data.hdf5 --cameras agentview_1 robot_wrist
python scripts/visualize_hdf5_open3d.py data.hdf5 --info
```

Keyboard controls in the Open3D window:

| Key | Action |
|-----|--------|
| Right / `N` | Next step |
| Left / `P` | Previous step |
| `R` | Reset to step 0 |
| Space | Toggle auto-play |

### `pyproject.toml`

Added `h5py` and `open3d` to the project dependencies and to the mypy `ignore_missing_imports` overrides.

---


pyproject.toml
```diff
@@ -32,6 +32,8 @@ dependencies = [
    "pybullet-arm64>=3.2.8",
    "tomsgeoms2d>=0.0.1",
    "pybullet_helpers>=0.1.0",
+   "pyqt5>=5.15.11",
+   "open3d>=0.19.0",
```
tomsilver: **Broken on a clean install + heavy deps in the wrong place.**

On a fresh `.[develop]` install, pylint collection fails with `ModuleNotFoundError: No module named 'h5py'`: `h5py` is imported (unguarded) by `demos_to_hdf5_with_pointclouds.py` and `visualize_hdf5_open3d.py`, but it's never declared as a dependency — it was only added to the mypy `ignore_missing_imports` list, which masks the problem under mypy while CI still breaks.

Meanwhile `open3d` and `pyqt5` *are* declared, but in the base `dependencies` where they don't belong — they're heavy native/binary wheels used only by the optional visualization scripts, not the core env library. (`open3d`'s import is already wrapped in a try/except with an install hint; `pyqt5` is never imported at all, only mentioned in error-message strings.)

Fix: declare all three in a dedicated optional extra (e.g. `pointcloud = ["h5py", "open3d", ...]`), consistent with the existing per-env extras pattern, and drop them from the base `dependencies`.

src/kinder/envs/dynamic3d/point_cloud.py
```diff
@@ -0,0 +1,413 @@
...
+        # pylint: disable=protected-access
+        robot_env = unwrapped._object_centric_env._robot_env
+    elif hasattr(unwrapped, "_robot_env"):
+        robot_env = unwrapped._robot_env  # pylint: disable=protected-access
+    else:
+        raise RuntimeError(
+            "Cannot locate a robot env inside the provided environment. "
+            "Expected either '_object_centric_env._robot_env' or '_robot_env' "
+            f"on {type(unwrapped).__name__}."
+        )
```
tomsilver, line 412: **The branch guard discriminates on the wrong attribute.** `_object_centric_env` is not dynamic3d-specific — it's assigned in the `core.py` base class (`ConstantObjectKinDEREnv.__init__`, ~L223), so `hasattr(unwrapped, "_object_centric_env")` is true for *every* kinder env (2D, kinematic3d, dynamic3d). The dynamic3d-specific marker is `_robot_env` (`envs.py:144`).

Consequence: pass a non-dynamic3d env and you enter this branch, then `unwrapped._object_centric_env._robot_env` raises a raw `AttributeError` — contradicting the docstring's promise of a `RuntimeError` with a helpful message. The friendly error below can essentially never fire for the realistic misuse case.

Also worth reconsidering more broadly: this reaches through three layers of private attributes across three modules with no public accessor, so any internal refactor silently breaks point-cloud generation. A public `sim` accessor on the env base would make this a one-liner. And the `elif hasattr(unwrapped, "_robot_env")` branch looks like dead/speculative code — no env instantiated via `kinder.make` exposes `_robot_env` directly on `unwrapped`.

src/kinder/envs/dynamic3d/point_cloud.py
```diff
@@ -0,0 +1,413 @@
...
+    return xyz_world, rgb_points
+
+
+# ---------------------------------------------------------------------------
+# MjSim-level API
+# ---------------------------------------------------------------------------
+
+
+def generate_scene_point_cloud(
+    sim: "MjSim",  # type: ignore[name-defined]  # kinder MjSim
```
tomsilver: These `# type: ignore[name-defined]` on `"MjSim"` (here and on `get_sim_from_env` at L373) are masking a genuinely missing import — `MjSim` is never imported into this module. Resolve it properly:

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from kinder.envs.dynamic3d.mujoco_utils import MjSim
```

Then drop the ignores and get real type checking on the `sim` argument.

Separately, the `mujoco`-related ignores are redundant and can just be deleted: `mujoco.*` is already in mypy's `ignore_missing_imports` (pyproject:113), so `import mujoco  # type: ignore` (L29) and every `# type: ignore[name-defined]` on `mujoco.MjModel`/`mujoco.MjData` are unnecessary — `mujoco_utils.py` imports and annotates these with no ignores. No new blanket override is needed.

scripts/visualize_hdf5_open3d.py
```diff
@@ -0,0 +1,484 @@
...
+else:
+    for _b in ["Qt5Agg", "Qt6Agg", "TkAgg", "GTK3Agg", "WXAgg"]:
+        try:
+            matplotlib.use(_b)
+            import matplotlib.pyplot as _plt  # pylint: disable=import-outside-toplevel,reimported
+            _plt.figure()
+            _plt.close("all")
+            break
+        except Exception:  # pylint: disable=broad-except
+            continue
```
tomsilver: This matplotlib backend-selection logic is duplicated in `visualize_saved_point_cloud.py` (`_select_backend`), and the two copies have **already drifted**:
- candidate lists differ — `visualize_saved_point_cloud.py` also tries `WebAgg`, this one doesn't;
- on total failure, `visualize_saved_point_cloud.py` falls back to `Agg` and prints a "install PyQt5" warning, while this one does neither — if every backend fails it silently leaves a broken/partial backend selected.

Extract a single shared helper (e.g. a small `scripts/_viz_backend.py` with `select_matplotlib_backend(no_show)`) and call it from both, so the behavior can't keep diverging.

scripts/visualize_hdf5_open3d.py
```diff
@@ -0,0 +1,484 @@
...
+
+from __future__ import annotations
+
+import argparse
+import sys
+import time
+from pathlib import Path
+from typing import Any
+
+import h5py  # type: ignore[import-untyped]  # pylint: disable=import-error
```
tomsilver: Two cleanups on this import once the dependency is sorted (see the `pyproject.toml` comment):
- `# type: ignore[import-untyped]` is **already redundant** — `h5py.*` is in mypy's `ignore_missing_imports`, so this can be removed now. (Note it's also inconsistent: the `h5py` import in `demos_to_hdf5_with_pointclouds.py:61` has no such ignore.)
- `# pylint: disable=import-error` can be removed once `h5py` is actually installed in CI. Since CI installs `.[develop]` via `.github/actions/setup/action.yml`, adding `h5py` to the `[develop]` extra is sufficient — no `ci.yml` change needed.

tomsilver (commented):
Light review — ran the point-cloud pipeline locally (26 tests pass, mypy clean, HDF5 export on SweepIntoDrawer3D-o5 works end-to-end). A few things to address, inline:

- **Blocker:** on a clean `.[develop]` install, pylint collection fails because `h5py` is imported but never declared as a dependency; the heavy `open3d`/`pyqt5` deps also shouldn't be in the base install. See the `pyproject.toml` comment.
- `get_sim_from_env` branches on `_object_centric_env`, which isn't dynamic3d-specific, so the friendly error path is effectively unreachable.
- Some type-ignores can be resolved properly (`MjSim` import) or are already redundant (`mujoco.*` / `h5py.*` are in the mypy override).
- The matplotlib backend-selection logic is duplicated across the two viz scripts and has already drifted.

Nothing major — mostly packaging + cleanup. Thanks!

pyproject.toml
yixuanhuang98: Fixed in cfd133a. Moved `open3d` and `pyqt5` out of base `dependencies` and into a new `pointcloud = ["h5py>=3", "open3d>=0.19.0", "pyqt5>=5.15.11"]` optional extra (consistent with the existing per-env extras pattern). Added `h5py>=3` to both `pointcloud` and `develop` so `.[develop]` installs it for CI — the `# pylint: disable=import-error` guards on the `h5py` imports are now gone too (274a3ca).

src/kinder/envs/dynamic3d/point_cloud.py
yixuanhuang98: Fixed in cfd133a. The function now navigates to the ObjectCentricRobotEnv layer first — via `_object_centric_env` if present on the wrapper, otherwise treating `unwrapped` as the OCE directly — and then checks `getattr(oce, "_robot_env", None)` as the Dynamic3D-specific discriminator. Non-Dynamic3D envs that lack `_robot_env` now get a clear `RuntimeError` as the docstring promises.

Good point on the public accessor — reaching through three layers of private attrs is fragile. I'll leave that for a follow-up since it's a cross-env change, but agree it's the right direction.

src/kinder/envs/dynamic3d/point_cloud.py
yixuanhuang98: Fixed in cfd133a. Added `from typing import TYPE_CHECKING` with `if TYPE_CHECKING: from kinder.envs.dynamic3d.mujoco_utils import MjSim`, then dropped all `# type: ignore[name-defined]` comments on `MjSim`, `mujoco.MjModel`, and `mujoco.MjData`, and the `# type: ignore` on `import mujoco`. `mypy` is clean.

scripts/visualize_hdf5_open3d.py
yixuanhuang98: Fixed in cfd133a. Extracted a shared `scripts/_viz_backend.py` with `select_backend(no_show)` covering all six candidates (`Qt5Agg`, `Qt6Agg`, `TkAgg`, `GTK3Agg`, `WXAgg`, `WebAgg`) plus an `Agg` fallback with a "install PyQt5" warning. Both `visualize_saved_point_cloud.py` and `visualize_hdf5_open3d.py` now call it; the inline logic in both scripts is gone.

scripts/visualize_hdf5_open3d.py
yixuanhuang98: Both cleanups done. `# type: ignore[import-untyped]` was removed in cfd133a. And now that `h5py>=3` is declared in the `develop` extra, `# pylint: disable=import-error` has been dropped from both `visualize_hdf5_open3d.py` and `demos_to_hdf5_with_pointclouds.py` in 274a3ca — no `ci.yml` change needed.

src/kinder/envs/dynamic3d/point_cloud.py
```diff
@@ -0,0 +1,420 @@
...
+import math
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any, Sequence
+
+import mujoco
+import numpy as np
+from numpy.typing import NDArray
+
+if TYPE_CHECKING:
+    from kinder.envs.dynamic3d.mujoco_utils import MjSim
```
tomsilver: Optional (non-blocking): the `TYPE_CHECKING` guard isn't necessary here — I tried a plain top-level `from kinder.envs.dynamic3d.mujoco_utils import MjSim` locally and mypy, pylint, and the tests all pass (no circular import, since `mujoco_utils` doesn't import this module). The plain import is a touch simpler, and there's no real load-cost difference given `point_cloud.py` already imports `mujoco` eagerly. If you do switch, drop `TYPE_CHECKING` from the `typing` import on L27 too. Fine to leave as-is.

tomsilver (approved):
Thanks for the quick turnaround — all five points from the last round are addressed and verified locally (mypy clean, pylint clean on the changed files, 26 tests pass, and the HDF5 export still runs end-to-end):

- deps moved out of the base install into a `pointcloud` extra + `h5py` in `[develop]` ✅
- `get_sim_from_env` now keys off `_robot_env` via `getattr` and raises a friendly error for non-Dynamic3D envs ✅
- redundant `mujoco`/`MjSim` type-ignores removed; `MjSim` imported properly ✅
- backend selection extracted into `scripts/_viz_backend.py` and used by both scripts ✅
- h5py type/pylint ignores dropped ✅

Approving. One tiny optional follow-up inline.

src/kinder/envs/dynamic3d/point_cloud.py
tomsilver: sorry Claude says this is optional but I think we should definitely change it, the "if TYPE_CHECKING" thing is ugly 😜

yixuanhuang98: Good call — confirmed no circular import (509d925). `TYPE_CHECKING` guard and the conditional import are gone; `MjSim` is now a plain top-level import alongside the other `mujoco_utils` types.

PR #105, opened 2026-05-27
Add mesh-based full point cloud generation for Dynamic3D environments

LJ1356:
## Summary

- Add `MeshPointCloud` dataclass (`xyz`, `geom_indices`, `geom_names`) with `filter_by_geom()` and `to_dict()` methods, complementing the existing `PointCloud`
- Add `get_scene_meshes(sim)` returning local-frame `trimesh.Trimesh` objects for every geom (mesh, box, cylinder, sphere, capsule types)
- Add `generate_full_point_cloud(sim, *, num_points_per_geom=500)` which samples surface points from each geom and transforms them to world frame — full coverage, no occlusion, no camera dependency
- Add unified `generate_point_cloud(sim, method="camera"|"mesh", ...)` dispatcher so callers can switch methods with a single argument
- Add `trimesh>=4.0.0` to `[pointcloud]` optional deps and mypy overrides
- 19 new tests covering all new public APIs

## Usage

```python
sim = get_sim_from_env(env)

# Camera-based (existing) — coloured, view-dependent
pc = generate_point_cloud(sim, method="camera")
# pc.xyz -> (N, 3) float32;  pc.rgb -> (N, 3) uint8

# Mesh-based (new) — full surface coverage, no occlusion
mpc = generate_point_cloud(sim, method="mesh", num_points_per_geom=500)
# mpc.xyz -> (N, 3) float32;  mpc.geom_names -> list[str]
```

## Implementation notes

- Internal `_get_scene_geoms(sim)` helper centralises trimesh import and returns `(geom_id, name, mesh)` tuples, eliminating both the duplicate mid-loop `import` and the fragile geom-ID string parsing that was in `generate_full_point_cloud`
- World-frame transform uses a pure-numpy quaternion→4×4 matrix (`_quat_to_mat4`), removing the per-geom `trimesh.transformations` call from the hot loop
- Capsule mesh generation delegates to `trimesh.creation.capsule` instead of a hand-rolled hemisphere splicing approach
- `filter_by_geom` now re-indexes `geom_indices` to 0 in the returned sub-cloud, so `sub.geom_names[sub.geom_indices[i]]` is always valid
- `MeshPointCloud.to_dict()` now includes `geom_names` as a string array, enabling lossless `np.savez` round-trips
- Mesh PC size is deterministic (`n_geoms × num_points_per_geom`), so the redundant re-pad guard in `_write_step` has been removed

## Test plan

- [x] All 45 point cloud tests pass (47 total, 2 skipped for visualization/save)
- [x] 19 new tests cover `MeshPointCloud`, `get_scene_meshes`, `generate_full_point_cloud`, the dispatcher (including default method), and `filter_by_geom` on non-first geoms
- [x] Full CI suite clean (pre-existing kinematic3d failures unaffected)


yixuanhuang98 (commented):
## Code Review — PR #105: Mesh-based full point cloud generation

### Test Results

All **43 tests pass** (2 visualization tests skipped as expected). The 17 new tests provide solid coverage of all new public APIs.

---

### Overview

This PR adds a mesh-based alternative to the existing camera-based point cloud pipeline. The approach is sound: sample surface points from trimesh geometry primitives, transform to world frame via MuJoCo sim state, and return a typed `MeshPointCloud` dataclass. The unified `generate_point_cloud(method=...)` dispatcher is a clean API decision.

---

### Issues & Suggestions

#### 1. `_require_trimesh()` called twice in `get_scene_meshes` and `generate_full_point_cloud`

**`point_cloud.py`** — `get_scene_meshes` calls `_require_trimesh()` at the top, then imports `trimesh` again mid-loop with `import trimesh as _trimesh`. The helper's purpose is to centralise the import/error, but the direct mid-loop import bypasses it:

```python
# Inside get_scene_meshes, in the mjGEOM_MESH branch:
import trimesh as _trimesh  # pylint: disable=import-outside-toplevel
```

This direct import should be replaced with the result of `_require_trimesh()` (or cache it in a local variable at the top of the function) so the error message stays consistent and the code path is unified.

#### 2. `_transform_from_pos_quat` needlessly re-imports trimesh on every call

`_transform_from_pos_quat` calls `_require_trimesh()` on every invocation — once per geom. Since `generate_full_point_cloud` calls it in a loop over all geoms, this adds repeated module-lookup overhead. `_require_trimesh()` should be called once at the top of `generate_full_point_cloud` (or `get_scene_meshes`), and the result passed down.

#### 3. Geom-ID parsing from string is fragile

`generate_full_point_cloud` recovers the geom ID by string-splitting the name key:

```python
geom_id = int(name.rsplit("_geom", 1)[1])
```

This works today, but it couples the loop body to the naming convention in `get_scene_meshes`. If a geom's raw name already contains `_geom` (e.g. `left_geom_pad`), `rsplit("_geom", 1)` still works correctly due to `maxsplit=1`, but it's a fragile implicit contract. A cleaner approach is to have `get_scene_meshes` return `dict[int, Any]` (keyed by geom_id) or a `list[tuple[int, str, Any]]`, and let callers use the integer ID directly.

#### 4. `_make_capsule_mesh` reimplements what trimesh already provides

`trimesh.creation.capsule(radius, height)` exists and generates a proper capsule mesh. The current custom implementation (concatenating a cylinder with manually split hemispheres) is more complex and harder to maintain. Consider replacing with:

```python
def _make_capsule_mesh(model, geom_id, **_):
    radius, half_height = model.geom_size[geom_id][:2]
    trimesh = _require_trimesh()
    return trimesh.creation.capsule(radius=radius, height=2 * half_height)
```

#### 5. `MeshPointCloud.to_dict` silently drops `geom_names`

`to_dict` serialises only `xyz` and `geom_indices`, omitting `geom_names`. This means a round-trip through `np.savez` / `MeshPointCloud(**np.load(...))` is not possible, and callers lose the ability to interpret indices after loading. Consider either documenting this limitation explicitly or including `geom_names` as a string array (e.g. `np.array(self.geom_names)`).

#### 6. `_write_step` re-pads even though `_capture_step` already pads

In `demos_to_hdf5_with_pointclouds.py`, `_capture_step` pads/subsamples to `max_pts` when `max_pts is not None`. Then `_write_step` checks `if len(xyz_m) != mesh_pc_size` and pads *again*. For the common case where `max_pts` is set, the second pad is always a no-op but adds a confusing double-check. Either remove the re-pad in `_write_step` (trusting that `_capture_step` already sized correctly) or remove it from `_capture_step` and do it only once in `_write_step`, matching how the original camera path works.

#### 7. Missing test: `generate_point_cloud` default method

There is no test asserting that calling `generate_point_cloud(sim)` (no `method` arg) returns a `PointCloud`. The default is `method="camera"`, which is an important guarantee for callers relying on the documented default.

#### 8. Missing test: `MeshPointCloud.filter_by_geom` reindexes `geom_indices` to 0

`test_mesh_point_cloud_filter_by_geom` asserts `np.all(sub.geom_indices == 0)`, which tests that the returned sub-cloud re-normalises indices. But there is no test for filtering a geom that is **not** the first one (`geom_names[0]`) to confirm the mask and re-indexing both work for `index > 0`.

---

### Minor / Style

- `_capture_step`'s keyword-only args (`store_camera`, `store_mesh`, `num_points_per_geom`) use defaults that differ from the function's documented primary mode. Passing them positionally is already prevented (they are kw-only), but the defaults are never exercised in the call sites — every caller passes all three explicitly. This is fine, but worth noting that the defaults only matter for direct external use.

- The `mode_label` ternary in `main()` duplicates logic that is already expressed in the `--completepointcloud` / `--allpointcloud` mutually-exclusive group. A small `{(False,False):"camera only", (True,False):"mesh only", (False,True):"camera + mesh"}` lookup would be clearer and less error-prone.

---

### Summary

The core implementation is correct and well-structured. The main actionable items before merge are **#1** (duplicate import), **#3** (fragile geom-ID string parsing), and **#5** (silent loss of `geom_names` in serialisation). Items #2 and #4 are optimisation/simplification suggestions. Items #6–#8 are test coverage gaps.

yixuanhuang98 (approved):

PR #106, opened 2026-05-29
Add --trackpointcloud flag for point-to-point correspondence

LJ1356:
## Summary
  - Add `generate_track_point_cloud()` to `point_cloud.py` — samples fixed local-frame surface points once after `env.reset()`, then re-applies the current body+geom world
  transform each step so point index `i` always refers to the same surface location
  - Wire `--trackpointcloud` into `demos_to_hdf5_with_pointclouds.py`, storing `obs/track_pc_xyz` (N, total_pts, 3) float32 and `obs/track_pc_geom_indices` (N, total_pts) int32
  - Combinable with `--completepointcloud` and `--allpointcloud`; respects `--max-pts` and `--num-points-per-geom`

  ## Test plan
  - [ ] Run `--trackpointcloud` on a demo dir and verify HDF5 contains `track_pc_xyz` and `track_pc_geom_indices`
  - [ ] Verify inter-point distances are constant across timesteps (rigid body correspondence check)
  - [ ] Confirm combinable with `--completepointcloud` and `--allpointcloud`
  - [ ] CI passes


src/kinder/envs/dynamic3d/point_cloud.py
```diff
@@ -744,6 +744,85 @@ def generate_full_point_cloud(
...
+            name: mesh.sample(num_points_per_geom).astype(np.float32)
+            for name, mesh in base_meshes.items()
+        }
+
+    all_xyz: list[NDArray[np.float32]] = []
+    all_idx: list[NDArray[np.int32]] = []
+    geom_names: list[str] = []
+
+    for local_idx, (name, local_pts) in enumerate(local_points_dict.items()):
+        geom_id = int(name.rsplit("_geom", 1)[1])
```
yixuanhuang98: **Fragile name parsing — geom_id was already available.** `get_scene_meshes(sim)` is a thin wrapper around `_get_scene_geoms(sim)` that returns `{name: mesh for _, name, mesh in _get_scene_geoms(sim)}` — it drops the `geom_id`. Then this line has to reconstruct it by parsing the name string. If `_get_scene_geoms` ever changes its naming format, this silently breaks with a cryptic `ValueError` or `IndexError` at runtime.

The fix: use `_get_scene_geoms()` directly on the first call (exactly as `generate_full_point_cloud` does at L706), and store the geom_id alongside the sampled points — e.g. `dict[str, tuple[int, NDArray[np.float32]]]` (name → (geom_id, sampled_pts)). The geom_id is then available without any string parsing.

src/kinder/envs/dynamic3d/point_cloud.py
```diff
@@ -744,6 +744,85 @@ def generate_full_point_cloud(
...
+
+        world_transform = body_transform @ _transform_from_pos_quat(
+            model.geom_pos[geom_id], model.geom_quat[geom_id]
+        )
+
+        pts_h = np.hstack([local_pts, np.ones((len(local_pts), 1))])
+        xyz = (pts_h @ world_transform.T)[:, :3].astype(np.float32)
+
+        all_xyz.append(xyz)
+        all_idx.append(np.full(len(xyz), local_idx, dtype=np.int32))
```
yixuanhuang98, line 801: Minor doc gap: `geom_indices` here stores sequential loop-order integers (0, 1, 2…), not MuJoCo geom IDs. This is consistent with `generate_full_point_cloud`, but users of `--track-geom N` in the visualiser naturally expect N to be a MuJoCo geom ID. A one-liner in the docstring or the CLI `--track-geom` help would prevent confusion — e.g. *"geom indices are 0-based sequential positions as returned by `_get_scene_geoms()`, not MuJoCo's internal `geom_id`"*.

scripts/visualize_hdf5_open3d.py
```diff
@@ -420,8 +517,60 @@ def main() -> None:  # pylint: disable=too-many-locals,too-many-statements
...
+                    f"  Pre-loading {len(cameras)} camera streams "
+                    f"({total_mb:.0f} MB)...",
+                    flush=True,
+                )
+                for cam in cameras:
+                    _rgb_cache[cam] = obs_grp[f"{cam}_rgb"][:]
+                print("  Done.", flush=True)
+
+        # Mutable container so _load_pc can update point_colors before state exists
+        _pc_colors: list[NDArray[np.float64] | None] = [None]
```
yixuanhuang98: `_pc_colors: list[NDArray[np.float64] | None] = [None]` is a workaround for mutating a closure variable. The idiomatic Python pattern is `nonlocal`:

```python
_pc_colors: NDArray[np.float64] | None = None

def _load_pc(step: int) -> ...:
    nonlocal _pc_colors
    ...
    xyz, colors, _pc_colors = _load_track_step(
        raw_xyz, raw_idx, _pc_colors, args.max_pts, rng,
        track_geom=args.track_geom,
    )
```

Same spirit as Tom's note in PR #103: prefer the simpler, more readable form.

scripts/demos_to_hdf5_with_pointclouds.py
```diff
@@ -460,8 +510,20 @@ def _write_step(
...
+        idx_m = capture["mesh_pc_geom_indices"]
+        if len(xyz_m) != mesh_pc_size:
+            xyz_m, idx_m = _pad_or_subsample_mesh(xyz_m, idx_m, mesh_pc_size, rng)
+        obs_grp["mesh_pc_xyz"][step_idx] = xyz_m
+        obs_grp["mesh_pc_geom_indices"][step_idx] = idx_m
+
+    if store_track:
+        xyz_t = capture["track_pc_xyz"]
+        idx_t = capture["track_pc_geom_indices"]
+        if len(xyz_t) != track_pc_size:
```
yixuanhuang98: For track points, `len(xyz_t) != track_pc_size` is always `False`: the same `local_points_dict` is used every step (fixed geom count and `num_points_per_geom`), and `_capture_step` already pads/subsamples to `max_pts` before storing in `capture`. This branch is dead code for the track mode. If keeping it for defensive symmetry with the mesh path, a comment noting that it should never trigger would help; otherwise it can be removed.

yixuanhuang98 (changes requested):
Light review — mypy clean, pylint 10.00/10, 45 tests pass (no new tests added for the tracked point cloud path). Three things to address:

- **Blocker:** no tests for `generate_track_point_cloud`. The core correctness guarantee — that intra-geom pairwise distances are invariant across timesteps — should be automated, not just listed in an unchecked test plan. See the `test_point_cloud.py` inline comment.
- `generate_track_point_cloud` calls `get_scene_meshes()` then immediately re-parses each `geom_id` back out of the name string. That's fragile indirection; the geom_id was already available from `_get_scene_geoms()` before it was dropped. See the `point_cloud.py` inline comment.
- `_pc_colors = [None]` mutable-list closure is unusual; `nonlocal` is the idiomatic Python solution here. See the `visualize_hdf5_open3d.py` inline comment.

yixuanhuang98:
**Missing tests for `generate_track_point_cloud`** (`tests/envs/dynamic3d/test_point_cloud.py`, after L629)

The existing file has 45 tests covering the camera and mesh paths. No tests were added for the tracked point cloud. The core correctness guarantee — that intra-geom pairwise distances are invariant across timesteps (rigid body ⟹ isometry) — should be automated. A minimal sketch to add after the `generate_full_point_cloud` block:

```python
def test_generate_track_point_cloud_first_call_returns_local_dict(sim):
    pc, local_pts = generate_track_point_cloud(sim, None, num_points_per_geom=10)
    assert isinstance(pc, MeshPointCloud)
    assert local_pts is not None
    assert len(pc) > 0


def test_generate_track_point_cloud_rigid_correspondence(sim, env):
    """Intra-geom pairwise distances must be invariant across timesteps."""
    from scipy.spatial.distance import pdist  # noqa: PLC0415

    pc0, local_pts = generate_track_point_cloud(sim, None, num_points_per_geom=20)
    env.step(env.action_space.sample())
    pc1, _ = generate_track_point_cloud(sim, local_pts, num_points_per_geom=20)

    for name in pc0.geom_names:
        g0 = pc0.filter_by_geom(name).xyz
        g1 = pc1.filter_by_geom(name).xyz
        np.testing.assert_allclose(pdist(g0), pdist(g1), atol=1e-4,
                                   err_msg=f"distances changed for geom {name!r}")


def test_generate_track_point_cloud_consistent_size(sim, env):
    """Point count must be identical across calls (same local_points_dict)."""
    pc0, local_pts = generate_track_point_cloud(sim, None, num_points_per_geom=10)
    env.step(env.action_space.sample())
    pc1, _ = generate_track_point_cloud(sim, local_pts, num_points_per_geom=10)
    assert len(pc0) == len(pc1)
```

yixuanhuang98 (approved):

PR #107, opened 2026-06-01
(  Add delta EE actions and canonical point cloud transforms

LJ1356:
## Summary

  - Add `--canonicalpointcloud` flag: stores canonical per-geom surface points (sampled once after `env.reset()`) and per-timestep
   rigid-body transforms instead of full per-timestep point clouds
  - Add `actions_delta_ee_transform` (T, 4, 4): action-aligned EE displacement where `delta[t] = inv(ee_pose_before_action[t]) @
  ee_pose_after_action[t]`
  - Add `obs/ee_pose` (T, 4, 4): EE world pose at each step
  - Add `find_ee_frame()` to `point_cloud.py`: auto-detects EE frame from site or body candidates; supports explicit `--ee-site`
  flag
  - Geom name metadata preserved via `attrs["geom_name"]` and `geom_names`/`geom_hdf5_keys` datasets

  ## HDF5 schema (canonical mode)

  data/demo_N/
    actions                          # (T, act_dim) float32 — unchanged
    actions_delta_ee_transform       # (T, 4, 4) float32 — new
    obs/
      state                          # unchanged
      ee_pose                        # (T, 4, 4) float32 — new
    canonical_pointcloud/
      /xyz                 # (P, 3) float32, local-frame canonical points
      geom_names                     # ordered original geom names
      geom_hdf5_keys                 # sanitized HDF5 keys
    geom_transforms/
                           # (T, 4, 4) float32, world transform per step
  
  Reconstruction: `world_pts = (T @ pts_h.T).T[:, :3]`

  ## Test plan
  
  - [x] CI passes
  - [x] Smoke-tested: no camera PC arrays in canonical mode, correct shapes and attrs
  - [x] `delta[t]` near-identity for no-motion steps
  - [x] Geom name metadata round-trips correctly


src/kinder/envs/dynamic3d/point_cloud.py
```diff
@@ -823,6 +823,199 @@ def generate_track_point_cloud(
...
+            name ends with ``_geomN`` where ``N`` is the geom index).
+
+    Returns:
+        Mapping from geom name to (4, 4) float32 homogeneous transform.
+    """
+    model = sim.model.mj_model
+    data = sim.data.mj_data
+    transforms: dict[str, NDArray[np.float32]] = {}
+    for name in geom_names:
+        geom_id = int(name.rsplit("_geom", 1)[1])
```
yixuanhuang98: Same fragile `rsplit("_geom", 1)[1]` name-parsing as flagged in [PR #106](https://github.com/Princeton-Robot-Planning-and-Learning/kindergarden/pull/106) for `generate_track_point_cloud` (L788). `geom_names` here come from `sample_canonical_points` → `get_scene_meshes()`, which wraps `_get_scene_geoms()` and drops the `geom_id`. So this function has to reconstruct it by parsing the name string.

Both functions should instead use `_get_scene_geoms()` directly (as `generate_full_point_cloud` does at L706) and carry the `geom_id` alongside the sampled points, eliminating string parsing entirely.

src/kinder/envs/dynamic3d/point_cloud.py
```diff
@@ -823,6 +823,199 @@ def generate_track_point_cloud(
...
+
+    raise RuntimeError(
+        "No EE frame found in model. "
+        f"Tried sites: {', '.join(_EE_SITE_CANDIDATES)}. "
+        f"Tried bodies: {', '.join(_EE_BODY_CANDIDATES)}. "
+        "Pass --ee-site with the correct site or body name."
+    )
+
+
+def find_ee_site(sim: MjSim, site_name: str | None = None) -> str:
```
yixuanhuang98: **Dead code.** The docstring says "backward-compatible wrapper" but `find_ee_site` was never in the codebase before this PR — there is nothing to be backward-compatible with. A `grep` across the entire repo finds zero callers outside the definition itself.

Remove it; callers should use `find_ee_frame` directly and handle the `(name, frame_type)` return.

src/kinder/envs/dynamic3d/point_cloud.py
```diff
@@ -823,6 +823,199 @@ def generate_track_point_cloud(
...
+    need to know whether the frame is a site or a body.
+    """
+    name, _ = find_ee_frame(sim, site_name)
+    return name
+
+
+def get_ee_pose(
+    sim: MjSim,
+    frame_name: str,
+    frame_type: str = "site",
```
yixuanhuang98: Minor: `frame_type: str` admits any string; only `"site"` and `"body"` are valid. `Literal` is already imported at L41, so this is a free improvement:

```python
def get_ee_pose(
    sim: MjSim,
    frame_name: str,
    frame_type: Literal["site", "body"] = "site",
) -> NDArray[np.float32]:
```

Same fix applies to the `ee_frame_type: str` parameters in `_create_canonical_datasets` and `find_ee_frame`'s return type annotation.

scripts/demos_to_hdf5_with_pointclouds.py
```diff
@@ -652,10 +783,34 @@ def _process_demo(
         track_pc_size=track_pc_size,
     )
 
+    # Collect canonical state at step 0 (after reset, before any action).
+    # ee_poses_all will have num_frames+1 entries: steps 0..num_frames-1 plus
+    # the terminal pose captured after executing actions[-1].
+    sim_canonical = get_sim_from_env(env) if store_canonical else None
```
yixuanhuang98: `get_sim_from_env(env)` was already called at L707 (inside `if store_canonical:`) and the result stored in `sim`. `sim_canonical` is a second handle to the **same** `MjSim` object — `get_sim_from_env` is not a factory, it unwraps the env and returns the live sim.

The duplication forces three wordy `if store_canonical and sim_canonical is not None and ee_frame_name is not None:` guards (L794, L805, L851) where `if store_canonical:` alone is sufficient (both `sim` and `ee_frame_name` are guaranteed non-None by the earlier block). Drop `sim_canonical` and reuse `sim`.

scripts/demos_to_hdf5_with_pointclouds.py
```diff
@@ -704,7 +877,11 @@ def main() -> None:
     if args.max_demos is not None:
         demo_files = demo_files[: args.max_demos]
 
-    store_camera = not args.completepointcloud
+    store_canonical = args.canonicalpointcloud
+    # In canonical mode the transforms replace per-timestep XYZ arrays, so
+    # camera point clouds are suppressed by default.  Camera RGB images are
+    # also skipped unless a camera-based mode is explicitly combined.
+    store_camera = not args.completepointcloud and not store_canonical
```
yixuanhuang98: **Silent camera suppression may surprise users.** The PR description and `--canonicalpointcloud` help text both say canonical mode is "combinable with --completepointcloud, --allpointcloud, and --trackpointcloud", but this line unconditionally drops camera RGB images whenever `--canonicalpointcloud` is set — even if the user passes `--allpointcloud` expecting camera + mesh + canonical in a single run.

If the intent is that canonical mode always suppresses cameras, the combinability claim should be corrected in the help text (and maybe a warning printed when `--allpointcloud --canonicalpointcloud` are combined). If cameras *should* be collectable alongside canonical data, the condition should be `not args.completepointcloud` only.

yixuanhuang98 (changes requested):
Light review — mypy clean, pylint 10.00/10, 45 existing tests pass (no new tests for the canonical path). Four things to address:

- **Blocker:** no tests for any of the new public API (`find_ee_frame`, `get_ee_pose`, `sample_canonical_points`, `get_geom_world_transforms`). The delta-EE and canonical-transform correctness guarantees should be automated, not just smoke-tested by hand.
- `get_geom_world_transforms` repeats the same fragile `rsplit("_geom", 1)` name-parsing flagged in PR #106. See inline.
- `find_ee_site` is labeled "backward-compatible" but has no callers anywhere in the repo — it's dead code. See inline.
- `sim_canonical` in `_process_demo` is a duplicate handle to the same `MjSim` object already held by `sim`, leading to three redundant `sim_canonical is not None` guards. See inline.

yixuanhuang98:
**Missing tests for new public API** (`tests/envs/dynamic3d/test_point_cloud.py`, after L629)

Four new public functions are added with no unit tests. At minimum these should be covered:

```python
# find_ee_frame
def test_find_ee_frame_auto_detects_site(sim):
    name, frame_type = find_ee_frame(sim)
    assert frame_type in ("site", "body")
    assert isinstance(name, str) and len(name) > 0

def test_find_ee_frame_explicit_valid(sim):
    name, frame_type = find_ee_frame(sim)  # auto to get a known-good name
    name2, ft2 = find_ee_frame(sim, name)
    assert name2 == name and ft2 == frame_type

def test_find_ee_frame_invalid_raises(sim):
    with pytest.raises(RuntimeError, match="not found"):
        find_ee_frame(sim, "nonexistent_site_xyz")

# get_ee_pose
def test_get_ee_pose_shape_and_dtype(sim):
    name, ft = find_ee_frame(sim)
    T = get_ee_pose(sim, name, ft)
    assert T.shape == (4, 4) and T.dtype == np.float32
    # last row must be [0,0,0,1]
    np.testing.assert_allclose(T[3], [0, 0, 0, 1], atol=1e-6)

def test_get_ee_pose_rotation_orthogonal(sim):
    name, ft = find_ee_frame(sim)
    R = get_ee_pose(sim, name, ft)[:3, :3]
    np.testing.assert_allclose(R @ R.T, np.eye(3), atol=1e-5)

# sample_canonical_points
def test_sample_canonical_points_returns_dict(sim):
    pts = sample_canonical_points(sim, num_points_per_geom=10)
    assert isinstance(pts, dict) and len(pts) > 0
    for name, arr in pts.items():
        assert arr.shape == (10, 3) and arr.dtype == np.float32

# get_geom_world_transforms + delta EE correctness
def test_get_geom_world_transforms_shape(sim):
    pts = sample_canonical_points(sim, num_points_per_geom=5)
    xforms = get_geom_world_transforms(sim, list(pts))
    for name, T in xforms.items():
        assert T.shape == (4, 4) and T.dtype == np.float32

def test_delta_ee_near_identity_for_no_motion(sim, env):
    """delta[t] should be near-identity when the robot doesn't move."""
    name, ft = find_ee_frame(sim)
    T0 = get_ee_pose(sim, name, ft)
    # zero action = no motion
    env.step(np.zeros(env.action_space.shape))
    T1 = get_ee_pose(sim, name, ft)
    delta = np.linalg.inv(T0.astype(np.float64)) @ T1.astype(np.float64)
    np.testing.assert_allclose(delta, np.eye(4), atol=1e-3)
```

LJ1356:
Thanks for the review — all points from the last round have been addressed:

- **Missing tests** — added 8 unit tests covering `find_ee_frame`, `get_ee_pose`, `sample_canonical_points`, `get_geom_world_transforms`, and delta-EE near-identity for no motion ✅
- **`rsplit("_geom", 1)` fragile parsing** — `sample_canonical_points` now uses `_get_scene_geoms` directly and returns `dict[str, tuple[int, NDArray]]` (carrying `geom_id`); `get_geom_world_transforms` takes `dict[str, int]` instead of `Sequence[str]` ✅
- **`find_ee_site` dead code** — removed ✅
- **`sim_canonical` duplicate handle** — removed; all three verbose guards simplified to `if store_canonical:`, reusing the existing `sim` ✅
- **`frame_type: Literal["site", "body"]`** — applied to `find_ee_frame` return type, `get_ee_pose`, `_create_canonical_datasets`, and the local `ee_frame_type` variable ✅
- **Silent camera suppression** — removed `and not store_canonical` from `store_camera` so canonical mode no longer drops cameras when combined with `--allpointcloud` ✅

Also resolved the merge conflict with main after PR #106 was merged.

src/kinder/envs/dynamic3d/point_cloud.py
LJ1356: Fixed in e72f1e4. `sample_canonical_points` now uses `_get_scene_geoms()` directly and returns `dict[str, tuple[int, NDArray[np.float32]]]`, carrying the `geom_id` alongside the sampled points. `get_geom_world_transforms` now takes `dict[str, int]` (name → geom_id), eliminating the string parsing entirely.

src/kinder/envs/dynamic3d/point_cloud.py
LJ1356: Fixed in e72f1e4. Removed entirely — `find_ee_site` was introduced in this PR with no prior callers anywhere in the repo, so there was nothing to be backward-compatible with. Callers should use `find_ee_frame` directly and unpack `(name, frame_type)`.

src/kinder/envs/dynamic3d/point_cloud.py
LJ1356: Fixed in e72f1e4. Changed to `Literal["site", "body"]` in `get_ee_pose`, `find_ee_frame`'s return type annotation, `_create_canonical_datasets`, and the local `ee_frame_type` variable in `_process_demo`.

scripts/demos_to_hdf5_with_pointclouds.py
LJ1356: Fixed in e72f1e4. Removed `sim_canonical`; all three guards simplified to `if store_canonical:` with `assert ee_frame_name is not None`, reusing the existing `sim`.

scripts/demos_to_hdf5_with_pointclouds.py
LJ1356: Fixed in e72f1e4. Removed `and not store_canonical` from `store_camera` — canonical mode no longer suppresses cameras when combined with `--allpointcloud`.

LJ1356:
@yixuanhuang98 re: missing tests — fixed in e72f1e4. Added 8 unit tests in `tests/envs/dynamic3d/test_point_cloud.py`:

- `test_find_ee_frame_auto_detects` — auto-detection returns a non-empty name and valid frame type
- `test_find_ee_frame_explicit_valid` — explicit name returns same result as auto-detect
- `test_find_ee_frame_invalid_raises` — unknown name raises `RuntimeError`
- `test_get_ee_pose_shape_and_dtype` — returns (4,4) float32 with unit last row
- `test_get_ee_pose_rotation_orthogonal` — 3×3 rotation block is orthogonal
- `test_sample_canonical_points_returns_dict` — returns non-empty dict of `(geom_id, local_pts)` per geom
- `test_get_geom_world_transforms_shape` — returns (4,4) float32 transform for each geom
- `test_delta_ee_near_identity_for_no_motion` — delta EE is near-identity when robot doesn't move

yixuanhuang98 (approved):

PR #108, opened 2026-06-03
 Add canonical point cloud visualization to visualize_hdf5_open3d.py

LJ1356:
- Add --canonicalpointcloud flag to reconstruct world-frame points from canonical_pointcloud/<geom>/xyz and geom_transforms/<geom>[step]
  - Add --geom, --geom-substr, --exclude-geom-substr for geom filtering
  - Add _load_canonical_geom_map, _filter_geoms, _load_canonical_step helpers
  - Print canonical point cloud info in --info mode )

### DIFF
--- a/scripts/visualize_hdf5_open3d.py
+++ b/scripts/visualize_hdf5_open3d.py
@@ -41,6 +41,14 @@ Usage
 
     # List available demos and cameras
     python scripts/visualize_hdf5_open3d.py data.hdf5 --info
+
+    # Visualise canonical point cloud, filtered to gripper geoms
+    python scripts/visualize_hdf5_open3d.py data.hdf5 \\
+        --canonicalpointcloud --geom-substr gripper --animate
+
+    # Visualise all canonical geoms at a specific step
+    python scripts/visualize_hdf5_open3d.py data.hdf5 \\
+        --canonicalpointcloud --step 50
 """
 
 from __future__ import annotations
@@ -433,12 +441,14 @@ def _load_canonical_geom_map(demo_grp: h5py.Group) -> dict[str, str]:
     Reads ``canonical_pointcloud/geom_names`` and ``geom_hdf5_keys`` when
     present; falls back to reading ``attrs["geom_name"]`` on each subgroup.
     """
+
+    def _dec(v: bytes | str) -> str:
+        return v.decode() if isinstance(v, bytes) else str(v)
+
     cp = demo_grp["canonical_pointcloud"]
     if "geom_names" in cp and "geom_hdf5_keys" in cp:
         names = cp["geom_names"][:]
         keys = cp["geom_hdf5_keys"][:]
-        def _dec(v: bytes | str) -> str:
-            return v.decode() if isinstance(v, bytes) else v
         return {_dec(n): _dec(k) for n, k in zip(names, keys)}
     # Fallback: iterate subgroups and read attrs
     result: dict[str, str] = {}
@@ -446,7 +456,7 @@ def _load_canonical_geom_map(demo_grp: h5py.Group) -> dict[str, str]:
         if k in ("geom_names", "geom_hdf5_keys"):
             continue
         orig = cp[k].attrs.get("geom_name", k)
-        result[str(orig)] = k
+        result[_dec(orig)] = k
     return result
 
 
@@ -539,6 +549,10 @@ def _load_canonical_step(
     all_xyz: list[NDArray[np.float64]] = []
     all_colors: list[NDArray[np.float64]] = []
 
+    # Per-geom cap so small geoms are not starved by large ones.
+    n_geoms = len(selected_geom_map)
+    per_geom_cap = max(1, max_pts // n_geoms) if n_geoms > 0 else max_pts
+
     for local_idx, (_orig, hdf5_key) in enumerate(selected_geom_map.items()):
         pts: NDArray[np.float32] = canon_pts[hdf5_key]  # (P, 3)
         T: NDArray[np.float32] = (
@@ -546,6 +560,9 @@ def _load_canonical_step(
         )
         pts_h = np.hstack([pts, np.ones((len(pts), 1), dtype=np.float32)])
         world_pts = (T @ pts_h.T).T[:, :3].astype(np.float64)
+        if len(world_pts) > per_geom_cap:
+            sel = rng.choice(len(world_pts), per_geom_cap, replace=False)
+            world_pts = world_pts[sel]
         color = np.array(cmap(local_idx % 20)[:3], dtype=np.float64)
         all_xyz.append(world_pts)
         all_colors.append(np.tile(color, (len(world_pts), 1)))
@@ -553,15 +570,7 @@ def _load_canonical_step(
     if not all_xyz:
         return np.empty((0, 3), dtype=np.float64), np.empty((0, 3), dtype=np.float64)
 
-    xyz = np.concatenate(all_xyz, axis=0)
-    colors = np.concatenate(all_colors, axis=0)
-
-    n = len(xyz)
-    if n > max_pts:
-        sel = rng.choice(n, max_pts, replace=False)
-        xyz, colors = xyz[sel], colors[sel]
-
-    return xyz, colors
+    return np.concatenate(all_xyz, axis=0), np.concatenate(all_colors, axis=0)
 
 
 # ---------------------------------------------------------------------------
@@ -683,6 +692,13 @@ def main() -> None:  # pylint: disable=too-many-locals,too-many-statements
                 file=sys.stderr,
             )
             sys.exit(1)
+        if args.canonicalpointcloud and "geom_transforms" not in demo_grp:
+            print(
+                "Error: geom_transforms not found in this demo. "
+                "Re-generate the HDF5 with --canonicalpointcloud.",
+                file=sys.stderr,
+            )
+            sys.exit(1)
 
         # ---- Resolve cameras (still used for the RGB panel) ---------------
         available_cams = _list_cameras(obs_grp)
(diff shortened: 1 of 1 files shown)


scripts/visualize_hdf5_open3d.py
```diff
@@ -477,6 +676,13 @@ def main() -> None:  # pylint: disable=too-many-locals,too-many-statements
                 file=sys.stderr,
             )
             sys.exit(1)
+        if args.canonicalpointcloud and "canonical_pointcloud" not in demo_grp:
+            print(
+                "Error: canonical_pointcloud not found in this demo. "
+                "Re-generate the HDF5 with --canonicalpointcloud.",
+                file=sys.stderr,
+            )
+            sys.exit(1)
```
yixuanhuang98, line 694: **Missing `geom_transforms` existence check — silent `KeyError` instead of a helpful message.**

The block above guards `canonical_pointcloud`, but line 755 does `demo_grp["geom_transforms"]` unconditionally. If a file has `canonical_pointcloud` but was written without `geom_transforms` (e.g. an older schema), the user sees a bare `KeyError: 'geom_transforms'` with no recovery hint. Add a guard that mirrors the pattern here:
```python
if args.canonicalpointcloud and "geom_transforms" not in demo_grp:
    print(
        "Error: geom_transforms not found in this demo. "
        "Re-generate the HDF5 with --canonicalpointcloud.",
        file=sys.stderr,
    )
    sys.exit(1)
```

scripts/visualize_hdf5_open3d.py
```diff
@@ -493,11 +699,20 @@ def main() -> None:  # pylint: disable=too-many-locals,too-many-statements
...
+            and not args.trackpointcloud
+            and not args.canonicalpointcloud
+        ):
             print("Error: no camera data found.", file=sys.stderr)
             sys.exit(1)
 
-        if args.trackpointcloud:
+        if args.canonicalpointcloud:
+            mode_label = "canonical PC"
+            if args.geom or args.geom_substr:
```
yixuanhuang98: **`mode_label` doesn't reflect `--exclude-geom-substr`-only filtering.**

```python
if args.geom or args.geom_substr:
    mode_label += " (filtered)"
```

If the user only passes `--exclude-geom-substr`, some geoms are excluded but the label still shows `canonical PC` rather than `canonical PC (filtered)`. Fix:
```python
if args.geom or args.geom_substr or args.exclude_geom_substr:
    mode_label += " (filtered)"
```

scripts/visualize_hdf5_open3d.py
```diff
@@ -517,10 +732,37 @@ def main() -> None:  # pylint: disable=too-many-locals,too-many-statements
...
 
-        # Pre-load all animation data into RAM to avoid per-frame HDF5 reads.
+        # ---- Canonical point cloud setup -----------------------------------
+        _selected_geom_map: dict[str, str] = {}  # orig_name -> hdf5_key
+        _canon_pts: dict[str, NDArray[np.float32]] = {}  # hdf5_key -> (P, 3)
+        _gt_grp: h5py.Group | None = None
+
+        if args.canonicalpointcloud:
+            try:
+                full_geom_map = _load_canonical_geom_map(demo_grp)
```
yixuanhuang98, line 759: **`full_geom_map` assigned inside `try` but read outside it — pyright/mypy will flag this as potentially unbound.**

Line 758 references `full_geom_map` after the `try/except` block. In practice it's safe (every exit from the `try` via the `except` also calls `sys.exit(1)`), but static analysers don't know that. Hoist the initialisation:
```python
full_geom_map: dict[str, str] = {}
try:
    full_geom_map = _load_canonical_geom_map(demo_grp)
    _selected_geom_map = _filter_geoms(...)
except ValueError as exc:
    ...
```
This makes the intent explicit: if execution reaches line 758, `full_geom_map` is always populated.

scripts/visualize_hdf5_open3d.py
```diff
@@ -365,6 +422,148 @@ def _load_track_step(
...
+    """Return a mapping from original geom name to sanitized HDF5 key.
+
+    Reads ``canonical_pointcloud/geom_names`` and ``geom_hdf5_keys`` when
+    present; falls back to reading ``attrs["geom_name"]`` on each subgroup.
+    """
+    cp = demo_grp["canonical_pointcloud"]
+    if "geom_names" in cp and "geom_hdf5_keys" in cp:
+        names = cp["geom_names"][:]
+        keys = cp["geom_hdf5_keys"][:]
+        def _dec(v: bytes | str) -> str:
```
yixuanhuang98: **`_dec` is invisible to the fallback path and causes a latent `bytes`-stringification bug.**

`_dec` is defined inside the `if` branch, so the fallback loop (lines 444-449) cannot use it. The fallback currently does `str(orig)` on line 449, which stringifies a `bytes` object as `b'name'` rather than decoding it to `'name'`. Hoisting `_dec` before the `if` fixes both:
```python
def _dec(v: bytes | str) -> str:
    return v.decode() if isinstance(v, bytes) else v

if "geom_names" in cp and "geom_hdf5_keys" in cp:
    ...
    return {_dec(n): _dec(k) for n, k in zip(names, keys)}
# fallback
for k in cp.keys():
    ...
    orig = cp[k].attrs.get("geom_name", k)
    result[_dec(orig)] = k  # was: str(orig)
```

scripts/visualize_hdf5_open3d.py
```diff
@@ -365,6 +422,148 @@ def _load_track_step(
...
+        all_colors.append(np.tile(color, (len(world_pts), 1)))
+
+    if not all_xyz:
+        return np.empty((0, 3), dtype=np.float64), np.empty((0, 3), dtype=np.float64)
+
+    xyz = np.concatenate(all_xyz, axis=0)
+    colors = np.concatenate(all_colors, axis=0)
+
+    n = len(xyz)
+    if n > max_pts:
```
yixuanhuang98: **Global uniform sub-sampling can silently drop small geoms entirely.**

After concatenating all geoms, a small geom with 50 points in a 200 K-point cloud has roughly a 0.025% chance of keeping each point, so it will almost certainly disappear while large geoms dominate. This is consistent with `_load_mesh_step` and `_load_track_step` so it is not a new regression, but since `--canonicalpointcloud` is designed for per-geom analysis, proportional per-geom sampling would give more balanced coverage:
```python
per_geom_cap = max_pts // len(selected_geom_map)
# cap each geom before concatenation instead of sampling globally
```

yixuanhuang98 (commented):
Overall the implementation is clean and consistent with the existing modes. Six issues below, ranging from a silent crash path to documentation gaps.

**Module docstring not updated.** The top-of-file docstring lists all major flags and usage examples but doesn't mention `--canonicalpointcloud`, `--geom`, `--geom-substr`, or `--exclude-geom-substr`. Consider adding a short example:
```
    # Visualise canonical point cloud, filtered to gripper geoms
    python scripts/visualize_hdf5_open3d.py data.hdf5 \\
        --canonicalpointcloud --geom-substr gripper --animate
```

LJ1356:
Fixed. Added two canonical examples to the module docstring:

```
    # Visualise canonical point cloud, filtered to gripper geoms
    python scripts/visualize_hdf5_open3d.py data.hdf5 \
        --canonicalpointcloud --geom-substr gripper --animate

    # Visualise all canonical geoms at a specific step
    python scripts/visualize_hdf5_open3d.py data.hdf5 \
        --canonicalpointcloud --step 50
```

scripts/visualize_hdf5_open3d.py
LJ1356: Fixed. Added a guard mirroring the `canonical_pointcloud` check:\n\n```python\nif args.canonicalpointcloud and "geom_transforms" not in demo_grp:\n    print(\n        "Error: geom_transforms not found in this demo. "\n        "Re-generate the HDF5 with --canonicalpointcloud.",\n        file=sys.stderr,\n    )\n    sys.exit(1)\n```

scripts/visualize_hdf5_open3d.py
LJ1356: Fixed. Extended the condition to cover all three filter flags:\n\n```python\nif args.geom or args.geom_substr or args.exclude_geom_substr:\n    mode_label += " (filtered)"\n```

scripts/visualize_hdf5_open3d.py
LJ1356: Fixed. Hoisted the initialisation before the `try` block:\n\n```python\nfull_geom_map: dict[str, str] = {}\ntry:\n    full_geom_map = _load_canonical_geom_map(demo_grp)\n    ...\nexcept ValueError as exc:\n    ...\n```\n\nStatic analysers can now see it is always bound before use at line 758.

scripts/visualize_hdf5_open3d.py
LJ1356: Fixed. Hoisted `_dec` before the `if` so both the fast path and the fallback use it. Also changed `str(orig)` to `_dec(orig)` in the fallback so bytes attributes are decoded correctly instead of being stringified as `b'name'`:\n\n```python\ndef _dec(v: bytes | str) -> str:\n    return v.decode() if isinstance(v, bytes) else str(v)\n\ncp = demo_grp["canonical_pointcloud"]\nif "geom_names" in cp and "geom_hdf5_keys" in cp:\n    ...\n    return {_dec(n): _dec(k) for n, k in zip(names, keys)}\n# Fallback\n...\nresult[_dec(orig)] = k  # was: str(orig)\n```

scripts/visualize_hdf5_open3d.py
LJ1356: Fixed. Replaced global sub-sampling with per-geom capping. Each geom gets at most `max_pts // n_geoms` points before concatenation, guaranteeing proportional representation for small geoms:\n\n```python\nn_geoms = len(selected_geom_map)\nper_geom_cap = max(1, max_pts // n_geoms) if n_geoms > 0 else max_pts\n\nfor local_idx, (_orig, hdf5_key) in enumerate(selected_geom_map.items()):\n    ...\n    if len(world_pts) > per_geom_cap:\n        sel = rng.choice(len(world_pts), per_geom_cap, replace=False)\n        world_pts = world_pts[sel]\n    ...\n```\n\nThe final concatenation is returned directly without a second global sub-sample pass.

yixuanhuang98 (approved):

PR #114, opened 2026-06-27
Shelf3d 500demos

yixuanhuang98:


tomsilver:
These demos need to be added to the black list so they don't run in unit tests! (58 minutes! 😱 )

PR #115, opened 2026-07-12
Add pick-and-place tasks and desk-mounted Franka FR3 robot (Robotiq 2F-85)

LJ1356:
## Summary

Adds pick-and-place tasks for two robots, including a new desk-mounted **Franka FR3 arm with a Robotiq 2F-85 gripper** — the first non-TidyBot robot fully wired into the Dynamic3D stack.

### New robot: Franka FR3 + Robotiq 2F-85

- `models/franka_fr3/fr3.xml` — [mujoco_menagerie](https://github.com/google-deepmind/mujoco_menagerie) `franka_fr3` with the 2F-85 gripper subtree (same one TidyBot uses) grafted onto the flange `attachment_site`; the menagerie keyframe is removed because keyframes don't survive scene merging. FR3 meshes live in the shared `models/assets/fr3/`; 2F-85 meshes are reused from `models/assets/2f85/` (no duplication). Menagerie LICENSE vendored alongside.
- `FR3RobotEnv` (`robots/fr3_robot_env.py`) — fixed-base arm with a `mount_height` setting (read from the task config) so it can be mounted on a desk surface; 8-D action space (7 arm joint position targets/deltas + gripper). Arm and gripper use MuJoCo position actuators, so no torque controller is needed.
- `MujocoFR3RobotObjectType` in `object_types.py` (mount pose + joint positions/velocities + gripper).
- `ObjectCentricFranka3DEnv` / `Franka3DEnv` in `envs.py`, and `"fr3"` in the robot dispatch. Robot-instance settings from the task JSON are now forwarded to the robot class as kwargs.
- Registration (`_register_dynamic3d`) now reads the robot type from each task config instead of hardcoding TidyBot3D.
- Shared-code cleanups: `IndexedView` promoted from a method-local class in `TidyBotRobotEnv` to `robots/base.py`; the post-reset settling loop uses the env's own action-space shape instead of the hardcoded 11-D TidyBot shape (latent crash for any non-TidyBot robot).

### New tasks

https://github.com/user-attachments/assets/1107845a-2f41-491d-b6b0-b644d772d4d8

- `kinder/FrankaPickPlace3D-o1-v0` — DROID-style desk scene: FR3 mounted on a 0.75 m desk, picks a cube and places it in a goal region on the desk.
- `kinder/PickPlace3D-o1-v0` — TidyBot picks a cube from the ground and places it in a ground goal region.

## Testing

- `tests/envs/dynamic3d/test_franka3d_pickplace.py` (3 tests) and `tests/envs/dynamic3d/test_tidybot3d_pickplace.py` (2 tests): reset/step smoke tests, robot mount-pose state check, and goal detection via `set_state` teleport into the goal region.
- Full local CI: all 269 `tests/envs/dynamic3d/` tests pass (no regressions from the `IndexedView`/`envs.py` changes); mypy and pylint clean on all changed files.
- Visual verification: rendered `task_view` frames for both tasks (robot, object, and goal region framing).


src/kinder/__init__.py
```diff
@@ -431,9 +432,15 @@ def _register_dynamic3d() -> None:
                 # Go through variants for this task
                 if task_config.is_file():
                     config_name = task_config.stem
-                    robot = "TidyBot3D"
-                    # Note: we only support one robot at the moment
-                    # In the future, get robot from config.
+                    # Map the robot type declared in the task config to the
+                    # environment class implementing it.
+                    with open(task_config, "r", encoding="utf-8") as f:
+                        robot_key = next(iter(json.load(f)["robots"]))
```
yixuanhuang98: **New import-time hard dependency on every task JSON's `"robots"` key.**

Before this PR, `_register_dynamic3d()` never opened task config files — it just hardcoded `robot = "TidyBot3D"`. Now `json.load(f)["robots"]` + `next(iter(...))` runs for *every* file under `tasks_root` during `_register_dynamic3d()`, which executes at `import kinder` time (via the module-level registration call), and the result is looked up in the hardcoded `{"tidybot": "TidyBot3D", "fr3": "Franka3D", "rby1a": "RBY1A3D"}` dict.

A task config that's missing `"robots"`, has an empty `"robots"` dict, or names a robot key outside those three now raises `KeyError`/`StopIteration` **at package import time**, breaking every kinder environment, not just the offending task. All checked-in configs happen to satisfy this today, but there's no validation guarding it — worth wrapping in a `try/except` that raises a clear error naming the offending config file, so a malformed future task fails loudly and locally instead of taking down `import kinder`.

src/kinder/envs/dynamic3d/robots/fr3_robot_env.py
```diff
@@ -0,0 +1,279 @@
...
+        assert self.qpos is not None, "Arm qpos must be initialized first"
+        assert self.ctrl is not None, "Arm ctrl must be initialized first"
+
+        self.qpos["arm"][:] = self.HOME_QPOS
+        # Position actuators: hold the home configuration
+        self.ctrl["arm"][:] = self.HOME_QPOS
+        # Ctrl values: 0 = fully open, 255 = fully closed
+        self.ctrl["gripper"][:] = 0.0
+        self.sim.forward()  # Update the simulation state
+
```
yixuanhuang98, line 227: **Verbatim duplicate of `TidyBotRobotEnv._update_ctrl`.**

This method (including its docstring) is character-for-character identical to `_update_ctrl` in `tidybot_robot_env.py` — both iterate `self.ctrl.items()` and slice the action array by cumulative part length.

This PR already promotes `IndexedView` out of `tidybot_robot_env.py` into `robots/base.py` for exactly this kind of reason (shared robot logic shouldn't live copy-pasted in two subclasses). `_update_ctrl` looks like a good candidate for the same treatment — move it onto `RobotEnv` in `base.py` and delete both copies, rather than letting a third copy appear the next time a robot is added.

src/kinder/envs/dynamic3d/envs.py
```diff
@@ -151,6 +159,7 @@ def __init__(
             camera_height=self.config.camera_height,
             seed=seed if seed is not None else self.seed,
             show_viewer=self.config.show_viewer,
+            **robot_config,
```
yixuanhuang98, line 179: **Unnamespaced kwargs forwarding can collide with the env's own constructor args.**

`robot_config` (the task JSON's `"robots"][robot_type][robot_name]` dict) is splatted in here via `**robot_config`, into a call that already passes `name`, `control_frequency`, `act_delta`, `horizon`, `camera_names`, `camera_width`, `camera_height`, `seed`, and `show_viewer` as named kwargs.

Nothing stops a future task config's `"robot": {...}` block from using one of those same keys (e.g. `"horizon"` or `"seed"`) — that would raise `TypeError: got multiple values for keyword argument` at env construction. Only `FrankaPickPlace3D-o1.json`'s `mount_height` is exercised today, so this isn't triggered yet, but there's no validation/namespacing preventing a task-config author from picking a colliding key later. Worth a quick `assert not (robot_config.keys() & {...reserved names...})` or similar guard with a clear message.

yixuanhuang98 (commented):
Reviewed the FR3 robot integration and pick-and-place task additions (vendored mesh files excluded from review scope). Left a few per-line comments below.

LJ1356:
Thanks for the review — all four points addressed in 9341271:

- **Import-time hard dependency on task JSONs' `robots` key**: the lookup in `_register_dynamic3d()` is now wrapped in `try/except (JSONDecodeError, KeyError, StopIteration)` and re-raises a `RuntimeError` naming the offending config file and the valid robot keys, so a malformed task config fails loudly and locally instead of taking down `import kinder`.
- **`_update_ctrl` duplication**: moved onto `RobotEnv` in `robots/base.py`. All three per-robot copies deleted (RBY1A had a third, functionally identical copy).
- **Un-namespaced `**robot_config` forwarding**: added an assert that the task config's robot block doesn't use any of the nine reserved constructor kwarg names, with a message naming the task config file and the colliding keys.
- **Object-centric near-duplication**: factored the shared 7-joint-arm + gripper logic into `ObjectCentricRobotEnv` helpers (`_get_arm_and_gripper_pos_data`, `_get_arm_and_gripper_vel_data`, `_set_arm_and_gripper_state`). TidyBot and FR3 now keep only their genuinely different parts (mobile base vs. fixed mount); the FR3-specific need to also write arm ctrl targets is an explicit `set_arm_ctrl_targets=True` flag. RBY1A's stub (base-only features) is unchanged.

All 266 dynamic3d tests pass locally; mypy/pylint clean on the changed files.

yixuanhuang98 (approved):
LGTM!

PR #116, opened 2026-07-15
Policy agnostic env descriptions

merlerm:
Went through all envs manually and removed every instance I found where a string was giving hints not just about how the env works but how to solve it. This is to ensure any approaches using LLMs to solve the environments given their code can't cheat by already knowing what to do. Hopefully I caught everything but there may be leftovers.

One big thing is that there is still definitely some aspect of this in the code itself, e.g. the class name ("PushPullHook") already implies what to do, and similarly comments and variables can probably inform the agent (e.g. if stuff is mentioning "the hook" and "moving the button" it's already implying how to solve this). But this does not seem fixable to me, and it's a thin line. At least this removes all blatant examples from the descriptions and an agent has to discover this stuff on its own.

After doing this manually I also had claude check and it didn't find anything.


tomsilver (approved):
LGTM, thank you! After this is merged I will release a new version.

PR #117, opened 2026-07-15
Support fixed-base robots (FR3) in canonical HDF5 export; add FR3 pinch_site

LJ1356:
## Summary

Makes `scripts/demos_to_hdf5_with_pointclouds.py --canonicalpointcloud` work for **fixed-base robots** (e.g. the desk-mounted Franka FR3), not just the mobile TidyBot. Previously it hard-failed on the FR3 because of a TidyBot-only base-pose lookup and a hardcoded gripper index.

## Changes

**`scripts/demos_to_hdf5_with_pointclouds.py`**
- **Gripper column** now reads the last action dim (`action[t][-1]`) instead of the hardcoded TidyBot index `10`, so it works for the FR3's 8-dim action (arm 7 + gripper 1) as well as TidyBot's 11-dim (base 3 + arm 7 + gripper 1). Same `[0,1]` Robotiq scaling → the `1 - 2*g` flip is unchanged and **bit-identical** for TidyBot (`-1 == 10` there).
- **`_get_base_pose_xy_yaw`** returns `None` (instead of raising) when there is no TidyBot `_joint_x/_joint_y/_joint_th` triplet. When absent, `actions_delta_base_3d` and `actions_delta_ee_base_10d` (and their convention attrs) are **omitted entirely** rather than zero-filled. TidyBot output is unchanged.

**`src/kinder/envs/dynamic3d/models/franka_fr3/fr3.xml`**
- Adds a `pinch_site` at the fingertip grasp point (pad midpoint, `pos="0 0 0.112038"`), oriented to match the gen3/TidyBot `robot_pinch_site` convention (computed by anchoring to the shared Robotiq 2F-85 pad geometry). This lets `find_ee_frame` auto-detect the EE frame and gives FR3 and sweep data one shared gripper-frame convention. The site is massless/collisionless — no effect on physics or determinism.

## Why omit (not zero-fill) the base fields
The downstream 3dwm converter consumes `actions_delta_ee_world_euler_gripper_7d` + `obs/ee_pose`; it never reads `base_3d`/`ee_base_10d`. A fixed-base delta is always zero, so omitting is honest and costs nothing downstream.

## Verification
- **FR3** (`FrankaPickPlace3D-o1-test`): exports cleanly with auto-detected `robot_pinch_site`, `has_base=False`, no base fields; `actions_delta_ee_world_euler_gripper_7d` present; gripper = `1 - 2*action[:,-1]`; world translation matches an independent recompute.
- **TidyBot regression** (`SweepIntoDrawer3D-o5`): still writes `base_3d`/`ee_base_10d`; gripper column bit-identical.
- **Pinch site**: auto-detects; world position matches the pad midpoint to ~5e-8 m; z-axis aligns with the gripper approach axis.
- **No side effects**: the two changed helpers have no external callers; the only `pinch_site` consumer is the `find_ee_frame` candidate list; all site access in the codebase is name-based (no hardcoded site IDs), so the new site's ID shift affects nothing.
- `black` / `docformatter` / `isort` / `mypy` / `pylint` clean.

### DIFF
--- a/scripts/demos_to_hdf5_with_pointclouds.py
+++ b/scripts/demos_to_hdf5_with_pointclouds.py
@@ -305,12 +305,13 @@ def _delta_T_to_7d(delta_T: NDArray[np.float32]) -> NDArray[np.float32]:
     return np.concatenate([translation, quat])
 
 
-def _get_base_pose_xy_yaw(sim: Any) -> NDArray[np.float32]:
+def _get_base_pose_xy_yaw(sim: Any) -> NDArray[np.float32] | None:
     """Return [x, y, yaw] of the TidyBot mobile base from MuJoCo qpos.
 
     Groups joints by name prefix and requires a complete {prefix}_joint_x / _joint_y /
-    _joint_th triplet from the same prefix. Raises RuntimeError for non-TidyBot
-    environments.
+    _joint_th triplet from the same prefix. Returns None for fixed-base robots (e.g. the
+    desk-mounted FR3) that lack such a triplet, so callers can skip the base-relative
+    action datasets.
     """
     import mujoco  # pylint: disable=import-outside-toplevel  # type: ignore[import]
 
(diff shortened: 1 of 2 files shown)


yixuanhuang98 (approved):
LGTM!

PR #123, opened 2026-07-31
Fix cluttered retrieval initial-state sampling

merlerm:
Fix three ineffective initial-state checks in ClutteredRetrieval2D without changing shared collision semantics:

- keep the full rotated target region within world bounds
- prevent the target from starting in or overlapping its goal
- prevent the goal region from overlapping the robot

Adds one deterministic regression test for each case. The full kinematic-2D test suite passes, and a 2,000-seed sweep found no remaining occurrences.


tomsilver (approved):
LGTM!

PR #126, opened 2026-08-06
Move the Tossing3D-o1 bin back inside blocks_goal_region

joshnroy:
# Josh hand-written

So when I was doing my experiments, I was confused why the skill-oracle (should be 100% right) kept missing the box and I looked into it. It looks like a previous pr moved the bin in the json config, but not the goal region (which explains the issue).

Change this reverts: https://github.com/Princeton-Robot-Planning-and-Learning/kindergarden/pull/29/changes#diff-3d96503905f45d1feb565e3147a94fab8d9293eabe63d14440df7a65991e4291R11

Few questions:

1. Should I make a github issue for this?
2. Should I keep the test file? I wasn't sure what the best practices for this codebase are
3. Should I move the bin to match the goal or the goal to match the bin?
4. Why was the bin moved to start with?
5. Assertion: As shown in the diagram below, it seems like the bin region is technically randomized by 1mm and the goal region isn't, so their overlap is 99+% at all times regardless. Is this correct or am I misunderstanding?

# AI Summary

Tossing3D-o1's bin sits outside `blocks_goal_region`, so a cube tossed **into the bin scores a failure** and a cube landing on bare floor short of it **scores a success**. Moving `bin_init_region` back to its pre-#29 value of 2.0 fixes both directions. One number, plus a test.

## Question / goal

Does Tossing3D-o1 score the task it documents — "the robot must toss the object into a bin" — and if not, what is the smallest change that makes it do so?

## Background

`bin_init_region` and `blocks_goal_region` sit in the same `regions` block and look alike, but they are different kinds of thing. One is **sampled** to place an object; the other is **never sampled** and only ever tested against. Nothing checks that the two agree.

```mermaid
flowchart TD
    subgraph SRC["Tossing3D-o1.json regions"]
        BIR["bin_init_region<br/>x 2.0 .. 2.001<br/>SAMPLED - 1 mm wide"]
        BGR["blocks_goal_region<br/>x 1.9 .. 2.1<br/>STATIC - never sampled"]
        CIR["blocks_init_region<br/>x 0.5 .. 0.75, y -0.25 .. 0.25<br/>SAMPLED - 250 x 500 mm"]
    end

    BIR -->|"draw a pose"| CENTRE["bin_0 centre<br/>x = 2.0005"]
    CENTRE -->|"+/- length/2 = 0.15"| FOOT["bin footprint<br/>1.8505 .. 2.1505"]

    BGR -->|"+/- 0.05 per side"| BOX["goal box<br/>1.85 .. 2.15<br/>z 0 .. 0.15"]

    CIR -->|"draw a pose"| CUBE["cube_0 start<br/>genuinely varies per seed"]
    CUBE --> THROW["robot throws"]
    THROW --> REST["cube comes to rest"]

    REST --> CHECK["check_in_region<br/>is rest inside the goal box?"]
    BOX --> CHECK
    CHECK --> SCORE["_check_goals"]

    FOOT -. "NOTHING COMPARES THESE" .-> BOX

    class BIR,CIR sampled
    class BGR static
    class CENTRE,FOOT,BOX,CUBE,REST derived
    class CHECK,SCORE check

    classDef sampled fill:#dbeafe,stroke:#1d4ed8,stroke-width:2px,color:#0b1324
    classDef static fill:#fef3c7,stroke:#b45309,stroke-width:2px,color:#0b1324
    classDef derived fill:#e5e7eb,stroke:#374151,color:#0b1324
    classDef check fill:#ede9fe,stroke:#6d28d9,stroke-width:2px,color:#0b1324
```

**Only `blocks_init_region` is meaningfully random** — 250 × 500 mm, so the cube's start genuinely varies per seed. `bin_init_region` is a sampling range too, but it is **1 mm wide**, so the bin lands at 2.0005 ± 0.5 mm every time. The goal region is not sampled at all.

**So why did the two ever line up?** Because the numbers were authored to make them:

```
goal region raw half-extent   0.10
ground inflation, per side  + 0.05
                            ------
                              0.15  =  bin length / 2  =  0.3 / 2
```

With the bin's centre on the goal region's centre, the goal box and the bin's footprint are **the same box** — [1.85, 2.15] against [1.8505, 2.1505], agreeing to 0.5 mm, which is just the 1 mm sampling jitter. That is what makes `blocks_goal_region` a sound proxy for "in the bin".

**Nothing enforces it.** The dashed edge is the defect: the bin's placed footprint is never compared against the goal box. When #29 moved `bin_init_region` to 2.23 and left `blocks_goal_region` alone, the bin's footprint became [2.0805, 2.3805] while the goal box stayed [1.85, 2.15] — overlapping by only 70 mm, with the proxy now failing in **both** directions:

| bin x | bin footprint | goal box | cube in the bin | cube on floor, short of it |
| --- | --- | --- | --- | --- |
| 2.23 (`main`) | 2.0805 .. 2.3805 | 1.85 .. 2.15 | rests at 2.2308 → outside the box → **failure** | rests at 2.0268 → inside the box → **success** |
| 2.0 (this PR) | 1.8505 .. 2.1505 | 1.85 .. 2.15 | rests at 1.9902 → inside → success | outside → failure |

And because a region's site defaults to `rgba` alpha `0` (`objects/base.py:370`, `:900`), none of this is visible in any render — which is why the clips below shade the goal box in.

`bin_init_region` moved 2.0 → 2.23 in [#29, `Tossing3D-o1.json` line 11](https://github.com/Princeton-Robot-Planning-and-Learning/kindergarden/pull/29/changes#diff-3d96503905f45d1feb565e3147a94fab8d9293eabe63d14440df7a65991e4291R11), "merge final changes from prpl-mono" — a 75-file, +148/−99 bulk sync whose same hunk renames `robot_0` → `robot`. **`Tossing3D-o2.json` is the control**: touched by that same commit, it received only the rename and still has the bin at 2.0 beside a byte-identical `blocks_goal_region`. The two variants of one task now disagree about where the bin goes.

## Hypothesis

None — implementation task, not an experiment.

## Guidance given

Minimal injection: change the one number, do not reformat, do not touch `o2`. Verify the mis-scoring by measurement rather than by reading the geometry. Check first whether 2.23 was deliberate — if it was, raise a question instead of a fix.

## Methods

Looked for a deliberate rationale before changing anything. #29 has an empty body and no review or issue comments; no PR or issue in the repo's history mentions Tossing3D, the bin or the goal region; `2.23` appears in no other human-authored file; every branch carrying o1 = 2.23 leaves o2 = 2.0; and nothing anywhere reads the bin's position.

Then measured the scoring directly: built `ObjectCentricTidyBot3DEnv`, dropped the cube, stepped MuJoCo to rest, read `_check_goals()` — repeating with only the JSON's bin x differing.

Added `tests/envs/dynamic3d/test_tidybot3d_tossing.py`, following `test_tidybot3d_pickplace.py` — the same env class, the same task-config-driven goal test, and the template this file is written from. All three tests read the bin's actual pose from the state rather than hardcoding a coordinate. Two are behavioural; the third pins the invariant they rely on — that the bin's footprint covers the goal region — because single sampled points only catch large drift.

### Main
image

### After this PR
image

Note: the goal region is shaded for the purposes of this demo

| bin x | cube in bin scores | cube on floor short of it scores |
| --- | --- | --- |
| 2.23 (`main`) | **failure** | **success** |
| 2.0 (this PR) | success | failure |

Both directions are wrong at 2.23: the check is unsound *and* incomplete. The floor case matters because the goal box reaches down to z = 0.

The clips above are the same defect in motion. Same oracle, same seed; the resting **z** is the tell — `0.0249` is bare floor, `0.0444` is sitting in the bin.

New tests: **3/3** pass at 2.0, **0/3** at 2.23. Sweeping the bin's x shows the coverage test is the sensitive one — at 2.149, 2.10, 2.05 and even **2.003** the two behavioural tests still pass **2/3** while coverage alone fails. It catches a 3 mm drift, which is already the regime where a floor landing scores. Not flaky: across seeds 0-5 the sampled bin x spans 2.00005-2.00081, a 1 mm range against a 2 mm tolerance. Full `tests/envs/dynamic3d/`: **269 passed, 4 skipped**. `./run_autoformat.sh` clean, pylint 10.00/10, `mypy .` clean across 124 source files.

## Recommendation

Merge. The evidence that 2.23 was swept in rather than chosen is strong — o2 is an unchanged control, the region's half-extents line up with a bin at 2.0 (`0.10 + 0.05 = 0.15 = length/2`), and no discussion or dependent file anywhere justifies 2.23. If that reading is wrong and 2.23 *was* deliberate, then `blocks_goal_region` is what needs to move instead, and I would rather hear that than have this merged quietly.

Two follow-ups. **The committed `docs/envs/**/Tossing3D*.gif` assets go stale when this merges, and deliberately are not regenerated here.** Four of the seven are o1-derived and affected; the two `-o2` variants are not, and `assets/group_gifs/Tossing3D.gif` is not produced by the generator at all. I attempted the regeneration twice and abandoned it both times: every frame comes back ~43/255 darker and translated ~100 px, **including the `-o2` pair this PR does not touch**, under EGL *and* OSMesa alike. Rendering here is deterministic and nothing committed since explains it, so the difference is environmental — the baseline was last written by `69bbaa5` (#82) in April, on a different machine. A regeneration on my side would therefore ship a ~15 MB whole-repo re-shading that cannot be shown to come from the bin move, which is exactly what this PR is claiming. **Regenerating on the machine that produced the baseline would keep the diff clean.** Worth knowing either way: `generate_env_docs.py` decides what is stale via `inspect.getfile(env.unwrapped.__class__)`, so a task-JSON change never matches and the assets go stale silently — `--env Tossing3D` or `--force` is required, and no CI job or documented procedure checks this. Separately, `Tossing3D-o3.json` on the `new_assets` branch inherits the same mismatch.

---

# Update: this PR now does two things

Everything above this line is the PR as originally opened and is left unedited, including its numbers — all of which were measured against the 1 mm sampling range that was still in place at the time. This section is additive. Where a number above describes that range (the "1 mm wide" framing, the "2.0005 ± 0.5 mm", the "sampled bin x spans 2.00005-2.00081 across seeds 0-5"), it describes the *previous* state of this branch rather than its current state.

**Please read this PR as two changes, not one:**

1. **The revert.** `bin_init_region`'s x moves back from #29's `2.23` to `2.0`. This needs no justification beyond "#29 broke it", and it is what the whole section above argues for.
2. **The jitter pin.** `bin_init_region` additionally stops being a *range* at all. This is a separate change with its own argument, made after the PR was opened, and it is what this section is about.

The second change is not implied by the first, and a reviewer should be able to reject it without rejecting the revert.

## Question / goal

`bin_init_region` sampled the bin's placement over a 1 mm × 1 mm band. Is that jitter doing any work, and if not, can it be removed without disturbing anything already measured?

## Background

The section above establishes that only `blocks_init_region` is meaningfully random: it is 250 × 500 mm, so the cube's start genuinely varies per seed. `bin_init_region` is a sampling range too, but a 1 mm one — the bin is scenery, and its position is not a variable of the task. The jitter was never chosen as a randomisation; it is an artifact of the region being expressed as a range at all.

Its one visible effect was that the bin's footprint sat a slightly different distance off `blocks_goal_region` on every seed, which is what forced the coverage test to carry a tolerance.

## Hypothesis

None — implementation task, not an experiment.

## Guidance given

Remove the jitter with a zero-width range, pinned at **`2.0`**. **Leave z alone.** Do not touch `Tossing3D-o2.json` or `barrier_init_region`. Do not restate or recompute any number published before this section.

An earlier revision of this section pinned at `2.0005`, the midpoint of the range being removed. That was corrected: `2.0` is the value that was actually authored, the old range was one-sided jitter *on top of* `2.0` rather than a distribution centred on `2.0005`, and `2.0` is the centre of `blocks_goal_region` — which is precisely the property the section above argues the bin needs. Pinning at the midpoint preserved the mean of a distribution whose removal was the point, and invented a coordinate appearing nowhere in the file's history.

## Methods

`Tossing3D-o1.json`, one line — shown against `main`, so it carries both changes:

```diff
-            "ranges": [[2.23, -0.0005, 2.231, 0.0005]],
+            "ranges": [[2.0, 0.0, 2.0, 0.0]],
```

**The 4-value form is kept deliberately.** `sample_pose_in_region` derives `z_start, z_end = 0, 2 * ground_placement_threshold` for a 4-value range, so **z is still sampled exactly as before** — only x and y are pinned. Converting to the 6-value form would have pinned z too, and that was explicitly not wanted.

**Why removing the jitter costs nothing in the random stream**, all three points checked against `objects/base.py:sample_pose_in_region` rather than assumed:

1. **Zero width is legal.** Bounds are validated with `if x_start > x_end: raise` — strictly greater, so equal bounds pass. Same for y and z.
2. **The draws are unconditional.** `np_random.uniform` is called once per axis (x, y, z, yaw) with no branching, so collapsing a range changes neither the number nor the order of draws.
3. **A degenerate draw still consumes randomness.** `uniform(a, a)` computes `low + (high - low) * random_sample()`: it pulls a double, multiplies it by zero and returns `low`. It does not skip the draw.

Point 3 is the load-bearing one, so it was measured rather than reasoned about: over seeds 0-19 the bit-generator state after the bin's four draws, and the next three draws taken from it, are identical between the old range and the pinned value — **20/20 seeds on both counts**. A separate check confirms `uniform(a, a)` advances the stream by exactly one draw, and that `uniform(2.0, 2.0)` returns exactly `2.0` on 100/100 seeds.

Then the same thing end-to-end in the simulator: build the env, `reset(seed=...)`, read the sampled poses and the `_check_goals()` verdicts, seeds 0-9, with and without the jitter.

## Results

**Nothing downstream moved.** `cube_0`'s sampled start position is identical on **10/10 seeds** — the observable consequence of the seed stream being preserved, since the cube is drawn after the bin.

**No scoring verdict changed on any seed.** For all of seeds 0-9, with and without the jitter: goal-after-reset `false`, cube-resting-in-the-bin `success`, cube-on-the-floor-short-of-the-bin `failure`. **0/10 verdicts flipped.**

The bin now lands at exactly `(2.0, 0.0)` on **10/10 seeds**.

**The bin's footprint and the goal box now coincide.** In **x** the edges agree *exactly* — the footprint is `[1.85, 2.15]` against a goal box of `[1.85, 2.15]`, gap `0.0` at both edges, where before each edge was off by up to 1 mm depending on the seed.

In **y** there is a residual of `2.8e-17 m` at each edge, and it is worth stating plainly rather than rounding away: the footprint is `[-0.15, 0.15]` while the inflated goal box is `[-0.15000000000000002, 0.15000000000000002]`, because the region's `0.10` half-extent plus `0.05` of ground inflation evaluates to `0.15000000000000002` in double precision rather than `0.15`. So the computed area fraction is `0.9999999999999999`, not exactly `1.0`. That is floating-point representation, not a geometric gap — the discrepancy is ~28 femtometres, and it is a property of `blocks_goal_region` that was equally present in every number measured before this change.

| | with the 1 mm jitter | pinned at 2.0 |
| --- | --- | --- |
| bin x, seeds 0-9 | 2.000054 .. 2.000988 | 2.0 on 10/10 |
| bin y, seeds 0-9 | -0.000124 .. 0.000415 | 0.0 on 10/10 |
| x-edge gap vs goal box | up to ~1 mm, seed-dependent | 0.0, both edges |
| footprint / goal-box area overlap | 99.5917% .. 99.9463% | 0.9999999999999999 on 10/10 |
| scoring verdicts changed | — | 0/10 |
| `cube_0` start position | — | identical, 10/10 |

Tests, run against this branch with and without the change: this PR's own tests **3/3 → 3/3**; full `tests/` **387 passed, 2388 skipped → 387 passed, 2388 skipped**, unchanged. `black --check` and `isort --check-only` clean on the touched test file; pylint **10.00/10** on it, with the `no_np_random` plugin confirmed loaded by a probe that expects and gets `C9900` (a wrong `PYTHONPATH` makes pylint print `E0013`, silently drop the plugin, and still report 10.00/10).

`kinder/Tossing3D-o1-v0` and `kinder/Tossing3D-o2-v0` both still register, and all 97 env ids still load — Dynamic3D registration is filesystem-derived, so a malformed task JSON would take down every environment, not just this one.

The coverage test's tolerance is **unchanged at `0.002`**; no test needed it widened. Its explanatory comment is updated, since it previously attributed the tolerance to placement jitter that no longer exists.

## Recommendation

Merge. Taking the two changes separately: the revert stands on the argument above; the pin is a strict simplification on top of it — the bin is scenery, it now sits on the goal region's centre, and every number published before this section still stands because the random stream is bit-identical.

**`Tossing3D-o2.json` is deliberately not touched.** Its role in this PR is as the untouched control on `main` — it is the evidence that #29's move of o1 to 2.23 was swept in rather than chosen — and that argument is about o2's state on `main`, so it is unaffected by what o1 does here. Its bin therefore keeps the 1 mm jitter, and the two variants now differ in that respect. Pinning o2 as well is a reasonable follow-up, but it is a separate change to a file this PR has no other reason to open.

`barrier_init_region` has the same 1 mm × 1 mm shape and is **not** changed here either, for the same reason: out of scope for this PR.


yixuanhuang98 (commented):
LGTM! Could we include a video of the new tossing skill?

yixuanhuang98 (approved):
LGTM!

PR #131, opened 2026-08-07
Add render stride for fine-timestep env doc GIFs

Minseo10:
Random-action GIFs assume ~25 coarse steps. Torque-controlled environments run
thousands of steps at 1/240 s, which would produce an unusably long GIF.

Adds a `render_every` stride to both the class-level and variant-level GIF helpers,
and a `FINE_TIMESTEP_ENVS` table keyed by class name that supplies `(num_actions,
render_every)` per environment.

Default behavior is unchanged: environments not in the table keep `num_actions=25`
and `render_every=1`, so existing GIFs regenerate identically.

Prepares for the LimbRepositioning3D environment in the follow-up PR.


tomsilver (approved):
LGTM!

PR #132, opened 2026-08-07
Add LimbRepositioning3D environment

Minseo10:
Adds `LimbRepositioning3D`: a robot welded to a passive human limb drives it to a
goal joint configuration using only joint torques, as an assistive robot would when
helping someone move an arm or a leg.

Sixteen variants pair a scene (`isolated`, `human`, `wheelchair`, `bed`) with a limb
(`left-arm`, `right-arm`, `left-leg`, `right-leg`). Each carries its own robot
placement, initial limb configuration, goal, and camera.

### Notes for review

- Unlike the rest of Kinematic3D, this steps PyBullet **forward dynamics with torque
  control** rather than resetting joint positions. It still needs exactly the same
  backend dependencies and lives in the same package, so it registers under
  `Kinematic3D` rather than introducing a category that would duplicate it.
- Gravity is off by default, so the limb resists only through its inertia and muscle
  tone. Set `gravity` in the config to enable it.
- Resets perturb each limb joint within that limb's anatomical range around the
  variant's nominal configuration and reject draws that push the limb into the scene,
  then re-solve the grasp. So the initial state varies with the seed while staying
  collision free.
- The joint limit and muscle tone models are pluggable. `SpringMuscleToneModel` is
  explicitly marked unvalidated and is not used by any registered variant.

### Testing

`pytest tests/envs/kinematic3d/test_limbrepositioning3d.py` — 60 passed. Covers
observations, determinism, reset from an init state, torque actually moving the
passive limb, and that all sixteen variants build with a tight grasp and sample
initial states within the limb's joint ranges.

Ported from https://github.com/empriselab/limb-manipulation, with TidyBot replacing
the Franka Panda used there.


tomsilver (approved):
LGTM!

PR #120, opened 2026-07-28
Remove Ground3D and Motion3D kinematic3d environments

yixuanhuang98:
## Summary
- Remove `Ground3D` and `Motion3D`, dev artifacts under the `Kinematic3D` category (`BaseMotion3D` is kept).
- Deletes source, tests, docs pages/variants, and demo/GIF assets for both environments.
- Updates registration (`src/kinder/__init__.py`), the README quick-reference table, and `scripts/docs/make_grid_video.py`'s env list accordingly.
- `tests/envs/kinematic3d/test_realistic_bg.py` used `Motion3D-v0` to exercise the generic (non-Motion3D-specific) `realistic_bg` feature; swapped to `Table3D-o1-v0`.

## Test plan
- [x] `kinder.register_all_environments()` registers `Kinematic3D` without `Motion3D`/`Ground3D`, with `BaseMotion3D` intact.
- [x] `./run_ci_checks.sh` — autoformat, mypy, and pylint all clean. The pytest failures present are pre-existing and unrelated (no GPU/EGL for MuJoCo rendering, an IKFast compiler-flag incompatibility in this sandbox), touching unrelated envs (Shelf3D, TidyBot3D, etc.) — not Ground3D/Motion3D.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #135, opened 2026-08-10
Validate Kinematic3D action bounds and motion paths

merlerm:
## Summary

This PR makes the Kinematic3D transition contract enforceable and reusable:

- clip mobile-base deltas to `max_action_mag`, matching the arm-joint clipping and declared action space;
- add an optional `max_collision_check_step` configuration for validating the interpolated path between the current and proposed configurations;
- reject the complete motion when the robot or a held object collides at an intermediate configuration; and
- update planner-backed tests to subdivide base paths into actions that respect the configured bound.

The swept check is disabled by default. Individual environments can opt in at a resolution appropriate for their geometry.

## Motivation

Kinematic3D previously checked only the endpoint of each action. It also applied the three mobile-base deltas without clipping them, even though those dimensions are bounded by `max_action_mag` in the action space. A caller could therefore submit an out-of-range base displacement, and an otherwise in-range action could move the robot or a grasped object through geometry as long as the endpoint was collision-free.

Packing3D makes this especially visible because its sources and destination are close together. The trajectory below solves a three-part instance in the theoretical minimum of six actions. Dense replay of the accepted actions finds collisions at intermediate configurations.

Packing3D endpoint-only trajectory

This is a transition-validity issue rather than a reward issue. The sparse reward favors short solutions, but it should not make collision-crossing actions valid.

## Corrected behavior

The environment layers above this PR opt into the shared validator and retain sparse rewards. Their corrected trajectories use bounded, collision-valid actions:

| Packing3D | Obstruction3D |
| --- | --- |
| Packing3D corrected behavior | Obstruction3D corrected behavior |

The Packing3D recording uses the same seed as the six-action trajectory above and completes in 89 actions under #138. The Obstruction3D recording is the repository demo regenerated under #137: eight in-space actions, no rejected motions, and terminal success. These are feasibility demonstrations of the opt-in environment layers; this base PR supplies the reusable transition mechanism.

## Transition semantics

When `max_collision_check_step` is set, the environment subdivides the clipped action so that no interpolated coordinate changes by more than the configured amount. It checks the same robot and held-object collision conditions at each sample. If any sample is invalid, the robot and held object return to their starting poses.

Gripper commands retain their existing semantics: the open or close command is processed after the motion is accepted or reverted, so moving and operating the gripper in one action remains supported.

## Compatibility

- Existing environments keep endpoint-only checking unless they opt in.
- Rewards and termination conditions are unchanged.
- In-bound endpoint-valid actions behave as before when swept checking is disabled.

## Test plan

- Shared-layer Kinematic3D regression suite: **16 passed**.
- Exact CI-style pylint check for `base_env.py`: **passed**.
- Added a regression that verifies an oversized base command is clipped to the configured bound.
- Added a regression that reaches a collision-free endpoint through an invalid path and verifies that swept validation rejects it.
- Updated planner-derived test trajectories so each base action respects `max_action_mag`.

## Stack

This is the bottom layer of native stack #139.

### DIFF
--- a/src/kinder/envs/kinematic3d/base_env.py
+++ b/src/kinder/envs/kinematic3d/base_env.py
@@ -80,8 +80,7 @@ class Kinematic3DEnvConfig(KinDEREnvConfig):
     end_effector_viz_half_extents: tuple[float, float, float] = (0.01, 0.01, 0.035)
     end_effector_viz_color: tuple[float, float, float, float] = (1.0, 0.2, 0.2, 0.0)
     max_action_mag: float = 0.4
-    # Maximum coordinate increment between swept-motion collision checks. When unset,
-    # only the action endpoint is checked, preserving the original transition.
+    # None disables intermediate collision checks.
     max_collision_check_step: float | None = None
     check_base_collisions: bool = False
 
@@ -527,9 +526,6 @@ class ObjectCentricKinematic3DRobotEnv(
             self.config.max_action_mag,
         )
         current_base_pose = self.robot.get_base()
-        next_base_pose = current_base_pose + SE2Pose(
-            base_action[0], base_action[1], base_action[2]
-        )
 
         # Store the current robot joints because we may need to revert in collision.
         current_joints = remove_fingers_from_extended_joints(
@@ -551,10 +547,7 @@ class ObjectCentricKinematic3DRobotEnv(
             self._robot_arm.joint_lower_limits[:7],
             self._robot_arm.joint_upper_limits[:7],
         ).tolist()
-        # Apply the motion incrementally so that an action cannot tunnel through a
-        # collision and end at a collision-free configuration. This remains a purely
-        # kinematic transition: intermediate states are used only for validity
-        # checking, with no physics integration or contact response.
+        # Validate the action path at evenly spaced configurations.
         max_delta = max(
             np.max(np.abs(base_action)),
             np.max(np.abs(np.asarray(next_joints) - np.asarray(current_joints))),
--- a/src/kinder/envs/kinematic3d/base_env.py
+++ b/src/kinder/envs/kinematic3d/base_env.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import abc
+import math
 from dataclasses import dataclass, field
 from typing import Any, Generic
 from typing import Type as TypingType
@@ -79,6 +80,9 @@ class Kinematic3DEnvConfig(KinDEREnvConfig):
     end_effector_viz_half_extents: tuple[float, float, float] = (0.01, 0.01, 0.035)
     end_effector_viz_color: tuple[float, float, float, float] = (1.0, 0.2, 0.2, 0.0)
     max_action_mag: float = 0.4
+    # Maximum coordinate increment between swept-motion collision checks. When unset,
+    # only the action endpoint is checked, preserving the original transition.
+    max_collision_check_step: float | None = None
     check_base_collisions: bool = False
 
     # This is used to check whether a grasped object can be placed on a surface.
@@ -517,7 +521,11 @@ class ObjectCentricKinematic3DRobotEnv(
 
     def step(self, action: Array) -> tuple[_ObsType, float, bool, bool, dict]:
         # execute the base action
-        base_action = action[:3]
+        base_action = np.clip(
+            action[:3],
+            -self.config.max_action_mag,
+            self.config.max_action_mag,
+        )
         current_base_pose = self.robot.get_base()
         next_base_pose = current_base_pose + SE2Pose(
             base_action[0], base_action[1], base_action[2]
@@ -543,13 +551,42 @@ class ObjectCentricKinematic3DRobotEnv(
             self._robot_arm.joint_lower_limits[:7],
             self._robot_arm.joint_upper_limits[:7],
         ).tolist()
-        self._set_robot_and_held_object(
-            next_base_pose, next_joints, current_finger_state
+        # Apply the motion incrementally so that an action cannot tunnel through a
+        # collision and end at a collision-free configuration. This remains a purely
+        # kinematic transition: intermediate states are used only for validity
+        # checking, with no physics integration or contact response.
+        max_delta = max(
+            np.max(np.abs(base_action)),
+            np.max(np.abs(np.asarray(next_joints) - np.asarray(current_joints))),
         )
+        if self.config.max_collision_check_step is None:
+            num_collision_checks = 1
+        else:
+            if self.config.max_collision_check_step <= 0:
+                raise ValueError("max_collision_check_step must be positive")
+            num_collision_checks = max(
+                1, math.ceil(max_delta / self.config.max_collision_check_step)
+            )
+        motion_valid = True
+        for i in range(1, num_collision_checks + 1):
+            t = i / num_collision_checks
+            intermediate_base_pose = current_base_pose + SE2Pose(
+                t * base_action[0], t * base_action[1], t * base_action[2]
+            )
+            intermediate_joints = (
+                np.asarray(current_joints)
+                + t * (np.asarray(next_joints) - np.asarray(current_joints))
+            ).tolist()
+            self._set_robot_and_held_object(
+                intermediate_base_pose,
+                intermediate_joints,
+                current_finger_state,
+            )
+            if self._robot_or_held_object_collision_exists():
+                motion_valid = False
+                break
 
-        # Check for collisions.
-        if self._robot_or_held_object_collision_exists():
-            # Revert!
+        if not motion_valid:
             self._set_robot_and_held_object(
                 current_base_pose, current_joints, current_finger_state
             )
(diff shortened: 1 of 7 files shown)


merlerm:
We ran kinder-bilevel-planning on all four Kinematic3D environments, plus the scripted (domain-specific) experts and PPO/SAC retraining, before and after this change: **A0** = pre-stack `421609e`, **A1** = this PR `4e9a443`, **A2** = #136's reset fix `9bf8d0f`. Every comparison uses identical episode seeds across revisions (5 outer seeds × 50 episodes per cell), with per-step traces recording the raw pre-clip action and state hashes before/after each step. Planner cells additionally pin `PYTHONHASHSEED` per outer seed (shared across revisions) so paired cells are deterministic rather than subject to per-process search-order noise.

### Bilevel planning (success rate)

| Environment | Skills | A0 | A1 | A2 | Paired traces, A0 vs A1 |
|---|---|---:|---:|---:|---|
| BaseMotion3D | maintained | 0.988 | 0.988 | — | **250/250 byte-identical** |
| Transport3D-o2 | maintained (old) | 0.284 | 0.280 | 0.308 | 226 identical; **16 expected clips**; 8 planner wall-clock residuals ¹ |
| Transport3D-o2 | **bound-compliant** | 0.300 | 0.300 | 0.324 | **250/250 byte-identical** |
| Shelf3D-o2 | maintained (old) | 0.81 | 0.21 | — | A0 succeeds via 12–21 out-of-bound actions/episode ² |
| Shelf3D-o2 | **bound-compliant** | **1.000** | **1.000** | — | **250/250 byte-identical** |
| PrplLab3D-o2 | maintained (old) | 1.000 | hangs ³ | — | A0 uses out-of-bound actions on a subset |
| PrplLab3D-o2 | **bound-compliant** | **1.000** | **1.000** | — | **250/250 byte-identical** |

"Byte-identical" means the full per-step trace files (raw pre-clip actions + state hashes before/after every step) are equal under `cmp` — possible because clipping is the identity on in-bound actions, so for a compliant policy A1 executes literally the same floats as A0. For A1 vs A2 (Transport), traces are identical on 233/250 episodes; the other **17 differ only at reset**, verified geometrically as the repaired table-overlap states (17/17 overlap under A1, 0/17 under A2; those episodes go from 0/17 to 6/17 solved).

### Scripted (domain-specific) experts (success rate)

| Environment | Expert | A0 | A1 | A2 | Paired traces, A0 vs A1 |
|---|---|---:|---:|---:|---|
| BaseMotion3D | maintained | 0.992 | 0.992 | — | byte-identical |
| Transport3D-o2 | maintained (old) | 0.452 | 0.448 | 0.472 | 15 expected clips; 1 success lost to clipping |
| Transport3D-o2 | **bound-compliant** | 0.412 | 0.412 | 0.436 | **byte-identical, 103 = 103 exactly** |

### RL retraining (mean eval return over 50 episodes, seed 301, released training configuration)

| Cell | A0 | A1 | A2 | Released anchor |
|---|---:|---:|---:|---:|
| PPO BaseMotion3D | −98.9 | −97.6 | — | −98.5 |
| SAC BaseMotion3D | −74.6 ⁴ | −57.8 ⁴ | — | −96.3 |
| PPO Transport3D-o2 | −1200 | −1200 | −1200 | −1200 |
| SAC Transport3D-o2 | −1200 | −1200 | −1200 | −1200 |

¹ The old Transport skills plan near the 60 s budget; wall-clock marginality is the one noise source hash pinning can't remove. All 8 residuals are method-side (the planner chose a different action from an identical state); zero unexplained transitions or resets anywhere.
² Old-Shelf rates reproduced exactly across two independent runs; the spread across outer seeds reflects episode difficulty, not noise.
³ Deterministic planning livelock in a pre-existing unbounded IK sampler (issue 1 below) — the hang *is* this cell's result; root-caused with a stack dump.
⁴ SAC (async vectorized envs) is measured non-reproducible: an exact same-config repeat of the A0 cell scored −56.4 vs −74.6 — a larger swing than any cross-revision difference — so SAC A0/A1 gaps are within same-revision variance.

### Reading the tables

**In-bound behavior is bit-for-bit unchanged.** Wherever the acting method respects the declared action space, A0 and A1 produce byte-identical trajectories: BaseMotion under both methods, and the bound-compliant skills/experts on Transport, Shelf3D and PrplLab3D (250/250 trace pairs each). Across ~1,750 paired episodes there are **zero** cases of an in-bound action producing a different transition and **zero** A0/A1 reset differences. RL is additionally shielded structurally — the training stack wraps environments in `ClipAction`, so agents never submit out-of-bound actions (PPO Transport training curves are byte-identical across all three revisions), and PPO BaseMotion reproduces the released anchor.

**Every behavioral difference is the PR working as intended.** On Transport with the old skills, the divergences are 16 episodes where the *identical* raw base command exceeds the 0.2 bound (A0 executed it oversized, A1 clips it). For A1 vs A2, exactly the 17 geometrically-invalid resets change — an invalid→valid initial-state correction, not a policy regression.

**Shelf3D and PrplLab3D were silently depending on the missing enforcement.** Their skills follow raw BiRRT waypoints without the constant-distance remap Transport's skills already apply, emitting base deltas up to 3× the bound — old-Shelf drops 0.81 → 0.21 under enforcement and old-PrplLab livelocks. Two small kinder-baselines fixes (separate PRs to follow) restore both to a deterministic 1.000, *better than the old skills achieved even on A0*: adding the missing remap plus closed-loop waypoint following, and giving their planning configs the `max_skill_horizon: 1000` override the Transport configs already carry (bound-compliant navigation takes more, smaller steps, so the global default of 100 truncates planning rollouts). Notably, the 0.2 bounds on exactly these environments were set deliberately in prpl-mono #287 ("setup smaller max_action_mag for shelf and tablebox env", collision-checking audit) — this PR completes that intent by enforcing them.

**The swept check is validated as well.** Dormant by default (no environment sets `max_collision_check_step`; the byte-identity above proves zero effect) and correct when enabled: replaying all 1,494 first-matrix episodes through check-enabled environments produced zero unexplained divergences, every rejection reverts state exactly, and a spot-checked rejection is a genuine ~0.8 mm arm-through-box penetration on the way to a collision-free endpoint. Resolution is a real decision: BaseMotion trajectories are 100% swept-valid at 1 mm, but at 5 mm every successful Transport BP episode contains at least one grazing action — supporting the per-environment opt-in used by #137/#138.

### Issues found along the way (all pre-existing — tracking issues to follow)

1. **`pybullet_helpers`: unbounded IKFast free-joint sampler can livelock planning.** `ikfast/utils.py::free_joints_generator` ends in `while True: yield rng.uniform(...)`; on an unreachable IK target a caller without its own budget spins forever, and outer planning timeouts never fire (checked only between iterations). Observed on PrplLab3D under A1: 28.5 CPU-minutes against a 60 s timeout, deterministic under pinned hash seeds, hot frame confirmed by stack dump. Fix: bound the generator or enforce `max_time` around its iteration.
2. **`pybullet_helpers` + `prpl_utils`: `ConcatTrajectory` raises at t == duration on Python 3.12+.** The class clips query time to a `sum()`-cached duration and walks segments with a naive fold; CPython 3.12's compensated `sum()` can exceed the fold by one ULP, so a boundary query falls through and raises `Time X exceeds duration X` with both numbers printing identically. Invisible on ≤3.11. It silently failed 2/250 expert episodes and, once constant-distance remapping made boundary evaluations common, 65/91 Shelf3D planning attempts. Standalone 13-float repro available. Fix: return the final segment's endpoint on fall-through.
3. **`kinder-rl`: released checkpoints aren't self-contained.** Agents train inside `NormalizeObservation`/`NormalizeReward` but `agent.save()` persists only network+optimizer state — the normalization statistics are lost, so released checkpoints can't reproduce the trained policies (which is why we retrained rather than frozen-evaluated). Fix: persist the statistics with a restore path.
4. **`kinder-rl`: checked-in configs don't match the released runs.** The released archives show `num_envs: 16` and per-env `max_episode_steps` (100/1200); the repo has `num_envs: 8` and a global 300. Training from the repo as-is fails to learn BaseMotion entirely (−300 floor vs released −98); with the archived settings we reproduce the release almost exactly. Fix: align configs or document the overrides.
5. **`kinder-baselines`: dead code for removed environments.** kindergarden #120 removed Ground3D and Motion3D, but kinder-baselines still carries their skills, env-models, configs and tests, which import the deleted modules and cannot load. The ground3d skills additionally contain the same missing-remap bug fixed here for shelf3d/prpl3d — one more reason to delete rather than resurrect. Suggested action: remove the orphaned modules.

### Bottom line

No unexplained regressions anywhere. Baselines that respect the declared action space are provably untouched — byte-identical trajectories, not merely matching scores. The two baselines that break were relying on the missing enforcement, and small accompanying kinder-baselines fixes restore both above their old strength under the new contract. From the regression-evidence side, #135 and #136 are good to merge.

tomsilver (approved):
Thanks a lot for doing this carefully!

merlerm:
@tomsilver sorry this took a while, finally got around to it. Claude wrote that response but I did go through and double check everything myself and it seems to make sense, still, maybe it would be better if someone else (maybe Yixuan) checked this before merging? Or are you happy enough with this?

Also opening a couple PRs with the fixes I did on kinder-baselines, i'm also happy to check the other issues claude found and see if they are worth small PRs.

merlerm:
> Thanks a lot for doing this carefully!

perfect, will merge for now then!

PR #136, opened 2026-08-10
Prevent Transport3D reset objects from overlapping the table

merlerm:
## Summary

This PR includes the table in Transport3D's movable-object collision checks during reset sampling. Both the transport box and source cubes are now rejected when their sampled geometry overlaps the table.

## Motivation

The reset bounds include space occupied by the fixed table, but the table was absent from the collision set used to validate sampled movable-object poses. This could initialize the box or a cube inside the table. Such states are invalid independently of the policy: they can be infeasible, or they can encourage one-step extraction motions that exploit the initial overlap.

The table is part of the environment's fixed obstacle geometry, so reset sampling should treat its occupied volume the same way it treats the other collision bodies.

## What changed

- Add the table body to the collision IDs used while sampling the transport box.
- Add the table body to the collision IDs used while sampling each cube.
- Add deterministic regressions for seeds that previously produced table overlap.

This PR does not change the reward, goal predicate, action scale, or nominal task distribution. It only removes invalid samples from that distribution.

## Test plan

- Focused Transport3D suite: **3 passed**.
- Regression seeds `57`, `97`, and `148` verify that neither the box nor any cube has positive-volume AABB overlap with the table after reset.

## Stack

Second layer of native stack #139. Depends on #135.

### DIFF
--- a/src/kinder/envs/kinematic3d/transport3d.py
+++ b/src/kinder/envs/kinematic3d/transport3d.py
@@ -280,7 +280,7 @@ class ObjectCentricTransport3DEnv(
             ),
             physics_client_id=self.physics_client_id,
             rng=self.np_random,
-            other_collision_ids={self.robot.base.robot_id},
+            other_collision_ids={self.robot.base.robot_id, self.table_id},
         )
 
         sample_collision_free_object_poses(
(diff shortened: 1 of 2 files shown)


tomsilver (approved):
LGTM!

PR #137, opened 2026-08-10
Require valid Obstruction3D placement and motion

merlerm:
## Summary

This PR tightens both parts of the Obstruction3D task contract:

- success requires the entire bottom footprint of the block to lie inside the target region at the correct resting height; and
- actions use a `0.2` maximum delta and opt into Kinematic3D swept collision validation at 5 mm spacing.

The environment retains its sparse reward of `-1` per nonterminal step. No dense shaping is added.

## Motivation

The previous goal test was based on support contact with the target body. It could accept a block beside the visible target, partially overhanging it, or at an incorrect height. Separately, the compact workspace, `0.4` action bound, and endpoint-only collision checking allowed generated controllers to finish in a handful of abrupt motions that crossed the target or obstructions between accepted endpoints.

Reducing the action bound alone improves visual granularity but does not make those paths collision-free. The corrected environment therefore combines an action scale consistent with the other small Kinematic3D manipulation environments and the shared swept validator.

## Behavior

| Previous generated controller, replayed unchanged | Regenerated collision-valid demo |
| --- | --- |
| Old controller rejected by corrected environment | Obstruction3D corrected behavior |

The left replay shows that the old shortcut-dependent action sequence no longer succeeds. The right recording is the repository's sole `Obstruction3D-o0` demo regenerated under the corrected transition: eight in-space actions, no rejected motions, and terminal success. It is a deterministic feasibility witness, not a distribution-wide policy evaluation.

## Goal predicate

Success now requires:

- the block is not grasped;
- all four bottom corners, transformed into the target-region frame, lie within its XY footprint; and
- the block's bottom is at the target's support height within the placement tolerance.

Checking corners in the target frame handles block yaw and rejects placements that only put the center inside the region.

## Test plan

- Focused Obstruction3D suite: **4 passed**.
- Changed-file pylint check: **passed**.
- Goal regressions cover centered and rotated valid placements, plus beside, overhanging, hovering, and distant invalid placements.
- Configuration regressions verify the `0.2` action bound and 5 mm swept-check spacing.
- The sole `Obstruction3D-o0` demo was regenerated under the corrected transition: 8 in-space actions, zero rejected motions, and both deterministic replay tests pass.
- Replaying the previously generated controller unchanged against the corrected environment rejects its invalid shortcut motions and no longer solves the episode.

## Stack

Third layer of native stack #139. Mechanically based on #136; its code changes otherwise depend only on the shared transition validator in #135.


tomsilver (approved):
LGTM but please test that the bilevel planning baseline still works the same on this one before and after

PR #138, opened 2026-08-10
Validate Packing3D motion, state, and geometry

merlerm:
## Summary

This PR makes Packing3D enforce its motion, state, and terminal-geometry contracts:

- use a `0.2` maximum action delta and opt into swept collision validation at 5 mm spacing;
- require every part to be fully seated inside the rack cavity, supported at the rack floor, and below the rim;
- reject strict overlap between seated parts;
- use analytic convex geometry for cuboids, right-triangular prisms, equilateral-triangular prisms, and their grasp pegs during reset sampling, carried-object collision checking, and terminal validation;
- reconstruct part types, dimensions, poses, and rack extents from the supplied state in `set_state()`; and
- raise `RuntimeError` if reset rejection sampling cannot find a valid part placement.

The reward remains sparse. This PR changes which transitions and terminal arrangements are valid, not how intermediate progress is scored.

## Motivation

The original three-part controller solved 100/100 evaluation episodes in 6.67 actions on average. On the example seed below it used exactly six actions: one jump-and-close and one jump-and-open per part. The endpoints were accepted, but dense replay found collisions along the interpolated robot and held-object paths. A `0.2` bound made the motion less abrupt but did not eliminate these invalid paths, so the environment also needs swept validation.

The terminal predicate had a separate geometry gap. It checked projected containment in the rack footprint, but did not establish that each part was resting at the rack floor, below the rim, and non-overlapping with every other part. PyBullet's fixed-link triangle mesh queries can also report a positive gap for visibly overlapping triangular parts, so the environment cannot rely on those queries alone for these shapes.

State restoration also needs to reproduce the geometry described by the supplied state. Randomized part and rack dimensions should not depend on the most recent reset, and triangle geometric centers must rotate their local centroid offset with the stored orientation.

## Relationship to #88

#88 fixed triangle drift by restoring the raw PyBullet body-origin pose and added a repeated no-drift regression. This PR keeps that behavior and test. Because full geometry reconstruction needs the same raw pose for parts and the rack, the raw-origin accessor now lives on `Packing3DObjectCentricState`; the redundant environment helper from #88 is removed. This layer adds orientation-aware geometric centers and reconstruction of state-defined types, dimensions, poses, body IDs, feature mappings, and rack extents.

## Behavior

| Before | After |
| --- | --- |
| Packing3D six-action endpoint-only solution | Packing3D corrected 0.2 solution |

Both recordings use seed `7138484576005690180`. The original trajectory terminates in six actions. The corrected trajectory uses 89 bounded, collision-valid actions and produces the separated terminal packing shown from above:

Packing3D corrected terminal state from above

## Geometry and state restoration

The overlap test represents each implemented part component as a convex polyhedron and applies the separating axis theorem over face normals and edge cross products. The grasp pegs are included as separate components. Exact contact is allowed; only positive-volume overlap beyond the numerical tolerance is rejected.

The same shape-aware predicate is reused in three places:

- reset rejection sampling;
- held-object collision validation during motion; and
- terminal validation between packed parts and between each part and the rack.

Triangle states retain their PyBullet body-origin pose for serialization. Their geometric-center accessor rotates the local centroid offset into the world frame. `set_state()` recreates cuboids, triangles, and the rack from the state's type and dimension features, rebuilds all body-ID mappings, and then applies the stored origin poses. A normal reset restores the configured rack geometry.

Reset rejection sampling now raises an explicit error after exhausting its attempt limit instead of retaining the final colliding sample.

## Goal predicate

Success requires every part to be ungrasped and fully seated in the rack cavity:

- its complete main-body footprint is contained in the rack's local XY bounds;
- its bottom is at the rack floor within the placement tolerance;
- its top remains below the rack rim;
- it has rack support without strict rack overlap; and
- no pair of seated parts strictly overlaps.

## Test plan

- Focused Packing3D suite: **21 passed**, sharded into three processes to avoid a load-related exit 137.
- All **46** Kinematic3D test nodes passed across branch-scoped shards.
- Changed-file pylint: **passed**; mypy on all four changed source files: **passed**.
- Pairwise overlap coverage includes every combination of cuboid, right-triangular, and equilateral-triangular parts.
- Regressions cover held triangular parts, peg-only overlap, exact contact, mixed-shape reset sampling, rack overlap, hovering, and valid seating.
- State regressions verify repeated no-drift restoration, exact round trips for rotated right triangles, and state-defined cuboid, triangle, and rack dimensions, including rebuilt ID and feature mappings.
- An infeasible-reset regression verifies that exhausted rejection sampling raises `RuntimeError`.

## Stack

Top layer of native stack #139. Mechanically based on #137; its code changes otherwise depend on #135 and merged PR #88.

### DIFF
--- a/src/kinder/envs/kinematic3d/packing3d.py
+++ b/src/kinder/envs/kinematic3d/packing3d.py
@@ -47,6 +47,7 @@ class Packing3DEnvConfig(Kinematic3DEnvConfig, metaclass=FinalConfigMeta):
     """Config for Packing3DEnv()."""
 
     # Robot.
+    max_action_mag: float = 0.2
     robot_base_home_pose: SE2Pose = SE2Pose(-0.12, 0, 0)
     robot_base_z: float = -0.4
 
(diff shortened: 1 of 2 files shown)


tomsilver (approved):
LGTM. That solution is wild!

PR #141, opened 2026-08-10
Add limb repositioning assets

Minseo10:
Assets for the LimbRepositioning3D environment, split out from the environment PR so
the Python changes can be reviewed without the binary payload.

- `assets/human/` — four 6-DOF limb URDFs, a torso, and their meshes
- `assets/wheelchair/`, `assets/hospital_bed/`, `assets/floor/` — scene furniture
- `light_blue_wall_texture.jpg` — room walls

Ported from https://github.com/empriselab/limb-manipulation, with TidyBot replacing the
Franka Panda used there.

No Python changes. These files are exercised by the tests in the limbs and scenes PRs
stacked above this one, which load every URDF and texture here; PyBullet raises on a
missing mesh, so a broken asset path fails those tests.

Part of a stack that splits the original LimbRepositioning3D PR into reviewable pieces.


tomsilver (approved):
Thanks for separating this out! It looks like these assets will add about 18 MB to the overall kindergarden package size. That's okay, but a note-to-self is that we should prioritize lazy download stuff like #104 in a near future PR.

PR #142, opened 2026-08-10
Add limb object types and torque-control utilities

Minseo10:
Object types for the robot, the passive limb, and static fixtures, plus the torque
action space, the wrap-aware joint distance, and a state class with convenience lookups
for limb and robot joints.

Torque control means the state carries joint velocities as well as positions, so these
types differ from the rest of Kinematic3D.

Tests: `tests/envs/kinematic3d/test_limb_utils.py` (14 tests) covers joint-distance
wrapping, the action space and its generated docs table, the GUI torque mapping, and the
state properties.

### DIFF
--- a/src/kinder/envs/dynamic3d/limb_object_types.py
+++ b/src/kinder/envs/dynamic3d/limb_object_types.py
@@ -1,0 +1,37 @@
+"""Object types for the limb repositioning environments."""
+
+from relational_structs import Type
+
+Limb3DEnvTypeFeatures: dict[Type, list[str]] = {}
+
+# Torque control means the state needs joint velocities as well as positions.
+Limb3DRobotType = Type("Limb3DRobot")
+Limb3DEnvTypeFeatures[Limb3DRobotType] = (
+    [
+        "pos_base_x",
+        "pos_base_y",
+        "pos_base_rot",
+    ]
+    + [f"joint_{i}" for i in range(1, 8)]
+    + [f"joint_vel_{i}" for i in range(1, 8)]
+)
+
+# A passive human limb (arm or leg) with 6 DOF.
+Limb3DLimbType = Type("Limb3DLimb")
+Limb3DEnvTypeFeatures[Limb3DLimbType] = (
+    [f"joint_{i}" for i in range(1, 7)]
+    + [f"joint_vel_{i}" for i in range(1, 7)]
+    + [f"goal_joint_{i}" for i in range(1, 7)]
+)
+
+# Fixtures are static scene objects with a pose, e.g., the wheelchair or the bed.
+Limb3DFixtureType = Type("Limb3DFixture")
+Limb3DEnvTypeFeatures[Limb3DFixtureType] = [
+    "pose_x",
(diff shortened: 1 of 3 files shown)


src/kinder/envs/kinematic3d/utils.py
```diff
@@ -417,3 +417,148 @@ def sample_collision_free_object_poses(
             raise RuntimeError(
                 f"Failed to sample collision-free pose for object {obj_id}"
             )
+
+
+JointTorques: TypeAlias = list[float]
+
+
+def joint_position_distance(positions: JointPositions, other: JointPositions) -> float:
```
tomsilver: This is a tricky thing that has caused bugs for us in the past! If joints are circular (wrap around), then this implementation looks correct, but sometimes joints have limits. There should be an implementation of this in `pybullet-helpers` that you can import and use. The key idea is that you need some `joint-infos` argument which indicates whether the joints have limits or are circular.

src/kinder/envs/kinematic3d/utils.py
```diff
@@ -417,3 +417,148 @@ def sample_collision_free_object_poses(
...
+
+
+ASSETS_DIR = Path(__file__).parent / "assets"
+HUMAN_ASSETS_DIR = ASSETS_DIR / "human"
+NUM_ROBOT_JOINTS = 7
+NUM_LIMB_JOINTS = 6
+PYBULLET_TIMESTEP = 1.0 / 240.0
+
+
+class LimbRepositioning3DObjectCentricState(ObjectCentricState):
```
tomsilver: Since these classes are specific to limb stuff, rather than shared utilities across all of kinematic3d, it may be better to put them in a file that is specific to limb stuff. I know those files don't exist as of this PR, but you could create them and add to them in the next PRs.

tomsilver (approved):
Looks great! Two small comments -- let me know if they are unclear. After addressing them, feel free to merge.

src/kinder/envs/kinematic3d/utils.py
Minseo10: https://github.com/Princeton-Robot-Planning-and-Learning/prpl-mono/blob/main/pybullet-helpers/src/pybullet_helpers/joint.py#L202
Is this the implementation? I'll fix it using this one in the next PR!

tomsilver: Thanks! That's the core difference method, but it returns a list of differences, one per joint. To turn that into an overall distance, there is a choice about how to combine over joints. See here: https://github.com/Princeton-Robot-Planning-and-Learning/prpl-mono/blob/919131377f42b277472e8a77a82b3819b581d95e/pybullet-helpers/src/pybullet_helpers/motion_planning.py#L390

src/kinder/envs/kinematic3d/utils.py
Minseo10: I moved them to limb_utils.py and created a new test file! Also modified `joint_position_distance`. You can check them at the latest commit

PR #143, opened 2026-08-10
Add passive human limb models

Minseo10:
PyBullet models of the human limbs the robot repositions, with the joint limit and
muscle tone models describing how they resist being moved.

These differ from the human models bundled with pybullet_helpers: some joint axes are
flipped, the joints are continuous, and the grasp frame is placed differently.

Tests: `tests/envs/kinematic3d/test_limbs.py` (33 tests) covers both joint-limit models,
both muscle-tone models, the factories and their error paths, the anatomical joint
ranges, and loading all four limbs in PyBullet.

### DIFF
--- a/src/kinder/envs/dynamic3d/limbs.py
+++ b/src/kinder/envs/dynamic3d/limbs.py
@@ -1,0 +1,307 @@
+"""PyBullet models of the human limbs that the robot repositions, and the joint limit
+and muscle tone models that describe how they resist being moved.
+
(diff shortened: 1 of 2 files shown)


tomsilver (approved):
LGTM! Re:
> These differ from the human models bundled with pybullet_helpers

This is likely to cause confusion -- we should probably just remove the human models from pybullet_helpers? I don't think anyone is actively using them.

PR #144, opened 2026-08-10
Add limb repositioning scenes

Minseo10:
The four scenes a limb can be repositioned in, in increasing order of clutter: isolated,
human, wheelchair, and bed.

`LimbRepositioningSceneConfig` derives the limb pose from the torso, or the torso pose
from the limb when `use_limb_pose_to_initialize` is set.

Tests: `tests/envs/kinematic3d/test_limb_scenes.py` (29 tests) covers the config
validation and both pose-derivation directions, the arm and leg grasp transforms, and
building all four scene types in PyBullet.

### DIFF
--- a/src/kinder/envs/dynamic3d/limb_scenes.py
+++ b/src/kinder/envs/dynamic3d/limb_scenes.py
@@ -1,0 +1,379 @@
+"""Scenes that a human limb can be repositioned in.
+
+There are four kinds of scene, in increasing order of clutter:
+
+* isolated: the limb floats in an empty world, with no body attached.
+* human: the limb is attached to a torso, along with the three other limbs.
+* wheelchair: a human seated in a wheelchair, in a room.
+* bed: a human lying on a hospital bed, in a room.
+"""
+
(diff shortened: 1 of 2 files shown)


src/kinder/envs/kinematic3d/limb_scenes.py
```diff
@@ -0,0 +1,379 @@
...
+    def get_limb_obstacle_ids(self) -> list[int]:
+        others = [
+            limb.robot_id
+            for name, limb in self.limbs.items()
+            if name != self.scene_config.limb_name
+        ]
+        return [self.torso_id] + others + super().get_limb_obstacle_ids()
+
+
+class _RoomScene(HumanScene):
```
tomsilver: Why is this one private? Just wondering

tomsilver (approved):
LGTM!

src/kinder/envs/kinematic3d/limb_scenes.py
Minseo10: Fixed it to public!

PR #145, opened 2026-08-10
Add the torque-controlled PyBullet simulation core

Minseo10:
## Motivation

This is the first of three PRs that #145 was split into, following the review. It carries the simulation core: config, physics client setup, scene and robot construction, grasp and weld, torque control preparation, `_apply_torques` and `_settle`.

## Changes

- Relocation from `kinematic3d/` to `dynamic3d/`
- **Torque was applied for one substep, not `num_substeps`.** PyBullet's `TORQUE_CONTROL` is not a persistent motor like `POSITION_CONTROL`; Bullet clears all applied multibody forces at the end of every `stepSimulation`. The old code issued the torque once and then stepped `N` times, so the first substep felt it and the rest coasted. It is invisible today because `dt == PYBULLET_TIMESTEP` gives `N == 1`, but `dt = 0.05` would have silently delivered a twelfth of the requested impulse. Torque is now re-issued on every substep.
- **Muscle tone moved into `_apply_torques` and is recomputed per substep**, since `get_muscle_tone_torque()` depends on the limb's current velocity. Subclasses add to it through the new optional `extra_limb_torque` argument rather than replacing it.
- `num_substeps` is `round(dt / PYBULLET_TIMESTEP)` with a `__post_init__` assertion that `dt` is a whole multiple of the timestep, replacing `int((dt + 1e-5) / PYBULLET_TIMESTEP)`.
- **Collision and damping settings now cover the base link.** Link indices run from `-1`, so both bodies previously kept their base link's default collision group, mask, and damping while the comment claimed otherwise.
- `enableJointForceTorqueSensor` removed: it was enabled on every joint of both bodies and never read.
- **`camera_pitch = 200.0` rendered the scene upside down.** PyBullet inverts the up vector outside `[-90, 90]`. The camera defaults now match `_FREE_CAMERA` in #132.
- `render_fps` is derived as `round(1 / dt)`; at `dt = 1/240` and `render_fps = 60`, saved videos played at four times real speed.
- `realistic_bg` is gone. It existed only so that `if "kinematic3d." in entrypoint` in three test files could inject it; that check no longer matches these environments.
- `scene: LimbRepositioningSceneConfig = None  # type: ignore` replaced with a real default, so `Limb3DEnvConfig()` constructs a usable scene instead of failing later inside `create_scene`.
- `_extend_joints_with_fingers` deleted in favour of the identical `extend_joints_to_include_fingers`.
- `_get_robot_joint_velocities` is annotated `JointVelocities` rather than `JointPositions`.

## Notes

`get_collision_clearance()` and `limb_penetration()` are removed here and reintroduced in the third PR of the split, where the bug in them is fixed and they are tested against a scene with obstacles.

## Tests

`pytest tests/envs/dynamic3d/test_base_limbrepositioning3d.py` — 17 passed. `mypy` and `pylint` clean on both files.

New tests cover the fixes that had none:

- `test_a_coarse_dt_applies_torque_for_the_whole_interval` compares one step at `dt = 4/240` against four steps at `dt = 1/240`. It fails on the old implementation by 1.08 rad.
- `test_config_validates_dt_and_derives_the_frame_rate` covers the `dt` assertion, the derived `render_fps`, and the default scene.

### DIFF
--- a/src/kinder/envs/dynamic3d/base_limbrepositioning3d.py
+++ b/src/kinder/envs/dynamic3d/base_limbrepositioning3d.py
@@ -1,0 +1,398 @@
+"""Base environment class for all limb repositioning environments.
+
+These are Kinematic3D environments, but use PyBullet forward dynamics with torque
+control, rather than the joint position resets used by the rest of the category.
+"""
+
+from __future__ import annotations
+
+import abc
+from dataclasses import dataclass
+from typing import Any, Generic, TypeVar
+
+import gymnasium
+import numpy as np
+import pybullet as p
+from numpy.typing import NDArray
+from pybullet_helpers.camera import capture_image
+from pybullet_helpers.geometry import SE2Pose, multiply_poses
+from pybullet_helpers.gui import create_gui_connection
+from pybullet_helpers.inverse_kinematics import (
+    InverseKinematicsError,
+    inverse_kinematics,
+    pybullet_inverse_kinematics,
+)
+from pybullet_helpers.joint import JointPositions
+from pybullet_helpers.robots import create_pybullet_mobile_robot
+from relational_structs import Array, ObjectCentricStateSpace, Type
+
+from kinder.core import KinDEREnvConfig, ObjectCentricKinDEREnv
+from kinder.envs.dynamic3d.limb_object_types import (
+    Limb3DEnvTypeFeatures,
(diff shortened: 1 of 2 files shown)


tomsilver (approved):
LGTM! 👏

PR #146, opened 2026-08-11
Add intermediate ConstrainedCupboard3D tasks

merlerm:
## Summary

Add registered `ConstrainedCupboard3D` task JSONs for object counts 3, 4, and 5, filling the gap between the existing `o2` and `o6` variants.

## Design

The fixture width now grows in two deliberate steps with the rod count:

- `o1` and `o2` retain the existing three-cupboard fixture;
- `o3` and `o4` use a centered six-cupboard fixture derived from the first six modules of `o6`;
- `o5` uses the same eleven-cupboard fixture as `o6`.

This avoids an abrupt geometry jump between `o5` and `o6` while keeping the number of distinct layouts small. Each added rod has its own shelf goal region and is present in both `initial_state` and `goal_state`.

Keeping these as explicit task JSONs follows the existing dynamic-3D registration model and avoids introducing a separate procedural layout convention.

## Initial-state distribution

The visualization uses 100 train seeds with counts `[1, 2, 3]` and 100 evaluation seeds cycling through counts `[1, 2, 3, 4, 5, 6]`. (train/test seeds are for a different project, here they are just meant to show the various configurations)

ConstrainedCupboard3D train and evaluation initial states

## Validation

- Added a parametrized test that creates each registered task, resets it, verifies the declared rod and cupboard counts, checks its observation, and constructs a state satisfying every goal predicate.
- Ran 100 seeded resets for each of counts 3, 4, and 5 after resizing the fixtures: all 300 resets succeeded, produced valid observations, and began outside the goal.
- Focused pytest: 3 passed.
- Focused mypy and pylint: passed.


tomsilver (approved):
We definitely should improve how these environments are parameterized but let's leave that for another day... (maybe never...)

PR #149, opened 2026-08-12
Resample constrained robot spawns in cluttered retrieval

merlerm:
## Summary

- resample cluttered retrieval initial states until the robot is in the largest collision-free component
- account for the robot base radius when computing free space
- add regression coverage for explicitly enclosed and dense sampled scenes

## Motivation

Collision-free sampling can place the robot in a small isolated pocket formed by the obstructions. The robot then starts severely constrained even though no individual object overlaps it.

Example of an enclosed robot spawn

The new reset-time geometry check rejects these small disconnected components without running an explicit planner or exposing a strategy-specific solvability check. It intentionally uses the largest free-space component regardless of whether that component touches the world boundary.

## Validation

- `uv run --extra develop python -m pytest -q tests/envs/kinematic2d/test_clutteredretrieval2d.py`
- 7 passed


tomsilver (approved):
Thanks! One comment

PR #153, opened 2026-08-13
Implement the environment contract in the limb base class

Minseo10:
## Motivation

Second of the three PRs that #145 was split into. Stacked on #145.

`ObjectCentricKinDEREnv` offers planners a "simulator as a model" interface:

```python
def get_transition(self, state, action):
    self._set_state(state)                              # put the sim into `state`
    obs, reward, terminated, _, _ = self.step(action)   # take one action
    return obs, reward, terminated
```

That is only truthful if the state carries everything the simulator needs and `_set_state` restores all of it. `Limb3DEnvTypeFeatures` gets this right — it carries `joint_vel_*` for both bodies, which is the correct call for a torque-controlled system — but the base class declared `_get_obs`, `_set_state` and `step` abstract, so the three methods that decide whether `get_transition` tells the truth lived outside the class that owns the simulator.

Two subclasses existed and one got it wrong. `LimbRepositioning3D` in #132 restores velocities correctly; the minimal subclass in #145's own tests dropped them, because `set_joints` defaults `joint_velocities` to zeros, and passed zero muscle tone. Nothing failed, because no test compared a predicted state against a stepped one.

## Changes

`_get_obs`, `_set_state` and `step` move into the base class. Subclasses keep only what genuinely differs between environments.

- `_set_state` restores joint positions **and velocities** for both bodies, then rebuilds the weld.
- `_get_obs` is built from `_object_names` and `_get_object_features`, both overridable, so a subclass with extra objects extends the list rather than reimplementing the state construction.
- `step` clips the action to the torque limits, applies it, and returns the observation, reward, termination and the distance-to-goal info dict.
- `_get_reward()` returns `-1.0` by default and `_get_extra_limb_torque()` returns `None`; subclasses override these instead of reimplementing `step`.
- `goal_reached` and `_reset_bodies` stay abstract.

The minimal test subclass drops from about fifty lines to sixteen.

## Notes

`_get_extra_limb_torque()` is the seam for the spasms in #132. That environment currently calls `_apply_torques` a second time to deliver a spasm, which advances the simulation by an extra `dt` with no robot torque and no muscle tone, making the control interval depend on a random draw. Overriding this hook instead folds the spasm into the same interval.

#132 should lose its `_set_state` entirely and shrink its `step` accordingly once this lands.

## Tests

`pytest tests/envs/dynamic3d/test_base_limbrepositioning3d.py` — 18 passed. `mypy` and `pylint` clean on both files.

`test_get_transition_matches_stepping` builds up real joint velocities over 100 steps, then checks that `get_next_state` agrees with actually stepping. This is the property the planning code depends on, and nothing checked it before.

### DIFF
--- a/src/kinder/envs/dynamic3d/base_limbrepositioning3d.py
+++ b/src/kinder/envs/dynamic3d/base_limbrepositioning3d.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 import abc
 from dataclasses import dataclass
-from typing import Any, Generic, TypeVar, cast
+from typing import Any, Generic, TypeVar
 
 import gymnasium
 import numpy as np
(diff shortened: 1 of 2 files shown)


tomsilver (approved):
LGTM!

PR #154, opened 2026-08-13
Add limb clearance and penetration measurement

Minseo10:
## Motivation

Third of the three PRs that #145 was split into. Stacked on the environment contract PR.

Both clearance functions were previously covered only by `test_isolated_scene_has_no_penetration`, and the isolated scene returns no obstacle ids at all. The loops never executed, `worst` kept its initial `0.0`, and the assertion passed — replacing the whole function body with `return 0.0` would not have failed that test. #144 provides bed and wheelchair scenes, so the measurement can now be checked against obstacles that actually exist.

## Changes

- `_CLEARANCE_QUERY_RADIUS = 0.5` replaces the magic number, and `_closest_distance` saturates at the radius instead of returning infinity when nothing is within range.
- `get_collision_clearance` starts its shrinking query radius at the constant rather than at infinity, so the first `getClosestPoints` call is given a finite distance.

The bug: `_closest_distance` returned `inf` when nothing was within 0.5 m, so an obstacle beyond that radius at construction time left `_limb_rest_clearance[body_id] == inf` forever. When the limb later came within range, the subtraction gave `0.4 - inf == -inf`, reporting infinite penetration for a limb that had touched nothing. Verified against the old code:

```
new _closest_distance(far body): 0.5
old _closest_distance(far body): inf
penetration if that were the baseline -> 0.4 - inf = -inf
```

This is where it would have done real damage. `_sample_limb_init_joint_positions` in #132 accepts a perturbed configuration only if `limb_penetration() > -init_max_penetration`. With `-inf`, every draw is rejected and the sampler silently falls back to the nominal configuration, producing an identical initial state for every seed while the docstring promises the opposite.

## Tests

`pytest tests/envs/dynamic3d/test_base_limbrepositioning3d.py` — 20 passed. `mypy` and `pylint` clean on both files.

Four tests, against a bed scene that has both limb obstacles and static bodies:

- `test_clearances_are_finite_at_rest` — both loops have bodies to run over, the rest clearances are bounded by the query radius, and the baseline is negative because the arm already overlaps the torso.
- `test_driving_the_limb_into_the_torso_reports_penetration` — swinging the shoulder inward reports penetration, even though collision response is disabled.
- `test_closest_distance_saturates_for_an_out_of_range_body` — the regression test for the fix above.
- `test_isolated_scene_saturates_rather_than_returning_infinity` — the degenerate case with no obstacles.

### DIFF
--- a/src/kinder/envs/dynamic3d/base_limbrepositioning3d.py
+++ b/src/kinder/envs/dynamic3d/base_limbrepositioning3d.py
@@ -50,6 +50,9 @@ from kinder.envs.dynamic3d.limbs import HumanLimbPyBulletRobot
 from kinder.envs.kinematic3d.utils import extend_joints_to_include_fingers
 from kinder.envs.utils import RobotActionSpace
 
+# Clearances are only queried within this radius, and saturate at it.
+_CLEARANCE_QUERY_RADIUS = 0.5
+
 # Default scene config for the limb repositioning environment.
 _DEFAULT_SCENE = LimbRepositioningSceneConfig(
     scene_type="isolated",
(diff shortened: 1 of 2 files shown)


tomsilver (approved):
LGTM!

PR #156, opened 2026-08-13
kinder: schedule control within a substep loop

joshnroy:
# Josh

In order to properly release a toss at the right time in Tossing3d, we need to have a high frequency controller, like we do on the real tidybot. This PR allows setting a sequence of actions rather than a single action to simulate that high frequency

# AI

## Summary

`MujocoEnv.step` held one action across every physics tick of a control period. At 10 Hz
control and a 0.5 ms physics timestep that is 200 ticks with a frozen command, so a
controller can only change what it asks for on a 100 ms boundary — long enough that a
gripper release, say, lands up to a tenth of a second late and by a different amount at
every commanded speed.

`step` now also accepts a **2-D action**: a schedule covering the control period exactly,
one row per `CONTROL_SCHEDULE_TIMESTEP`, with row `j` held for the `j`-th of them.
Nothing here knows what the rows mean — they are whatever this env's `_update_ctrl`
accepts — so every `Dynamic3D` env inherits it without any of them naming a gripper.

**Covering the period exactly is required, not merely allowed.** A shorter schedule would
need a rule for the ticks it omits, and any such rule reinterprets the caller's timing
without saying so. An earlier revision held the last row for the remainder, which made a
13-row and a 100-row schedule mean the same thing by accident; the consumer relied on
that, and now states its release index explicitly instead.

Three deliberate choices:

- **1 kHz, not the 2 kHz physics rate**, because that is the real Kinova arm's low-level
  servo period. In `jimmyyhwu/tidybot`'s `robot/kinova.py`,
  `MoveJController.CYCLIC_STEP_SIZE = 0.001` gates the send loop under
  `LOW_LEVEL_SERVOING` — one trajectory row per millisecond — and the same loop releases
  the gripper on `cyclic_count >= self.gripper_release_ms`. So on hardware the release
  parameter is *an index into the 1 kHz cycle counter*; matching that rate here is what
  makes a scheduled millisecond mean the millisecond the robot means. The same file's
  throw call passes `max_vel=140, gripper_release_ms=600`, which is also where the
  consumer's 140 deg/s ceiling comes from.
- **The schedule rides on the action, not on a new parameter.** `gymnasium.Wrapper.step`
  and `RenderCollection.step` both take exactly one positional argument and forward it by
  value, so a keyword or a second positional would not survive a wrapped env — and the
  recording path wraps every env it films. A test steps a schedule through
  `RenderCollection` and checks it reached the simulator.
- **The 1-D path allocates nothing.** A 1-D action becomes a read-only `np.broadcast_to`
  view over the period rather than a materialised array.
  `test_a_1d_action_is_held_for_every_tick_of_the_control_period` compares against an
  explicit tick-by-tick replay and asserts `np.array_equal` on `qpos` and `qvel`, rather
  than asserting closeness.

`TidyBotRobotEnv.step` gains the matching translation: its action-to-ctrl conversion is
split into `_action_to_ctrl` so each row of a schedule is converted the same way a lone
action is, against the state as it is *before* the period runs. A schedule re-commands
within a period; it does not re-close the loop on intra-period feedback.

**The consumer is `kinder-baselines` PR https://github.com/Princeton-Robot-Planning-and-Learning/kinder-baselines/pull/113**, which computes a toss's
release millisecond analytically and passes it here. That PR depends on this one; this
one stands alone.

**This one must merge first, and the dependency is harder than it reads.** The consumer
imports `CONTROL_SCHEDULE_TIMESTEP` from `kinder.envs.dynamic3d.mujoco_utils` at module
scope, so without this PR its `parameterized_skills` **fails to import at all** — a
downstream project that bumps its `kinder-baselines` pin without also bumping
`kindergarden` gets a deep `ImportError` behind a lazy import, far from the pin that
caused it. That happened downstream on 2026-08-13.

## Test plan

- [x] `python -m pytest tests/envs/dynamic3d/test_mujoco_utils.py -q` — **6 passed**
      (new file, 4 of those tests are this PR's)
- [x] `python -m pytest tests/envs/dynamic3d/test_tidybot3d_basic.py
      tests/envs/dynamic3d/test_tidybot3d_tossing.py
      tests/envs/dynamic3d/test_tidybot3d_ground.py
      tests/envs/dynamic3d/test_tidybot3d_utils.py -q` — 105 passed
- [x] `mypy src/kinder/envs/dynamic3d/mujoco_utils.py
      src/kinder/envs/dynamic3d/robots/tidybot_robot_env.py` — no issues in 2 source files
- [x] Downstream, at the consuming branch: `pytest kinder-models/tests/dynamic3d/tossing/
      -q` in `kinder-baselines` — **35 passed**

Formatting note: the changed files were formatted with `black` and `isort` at the repo's
own settings (line length 88, isort profile black). `run_autoformat.sh` was **not** run
repo-wide, deliberately — it currently rewrites ~28 unrelated files from tool-version
drift, and it never asserts a clean diff, so unformatted code does not fail CI anyway.

## Followup

- Only `TidyBotRobotEnv` translates a 2-D action today. `Fr3RobotEnv` and `Rby1aRobotEnv`
  override `step` with their own action parsing and would need the same one-line split
  before a schedule could reach them; neither has a caller that wants one yet.
- The schedule is uniform-in-time (one row per millisecond). A run-length form would be
  more compact for the common "hold, then switch once" case, but costs a second
  representation for no measured benefit at these sizes.

### DIFF
--- a/src/kinder/envs/dynamic3d/mujoco_utils.py
+++ b/src/kinder/envs/dynamic3d/mujoco_utils.py
@@ -26,6 +26,11 @@ from relational_structs import Array
 # to simulate for each step.
 SIMULATION_TIMESTEP = 0.0005  # (in seconds)
 
+# How long one row of a 2-D action is held for; see MujocoEnv.step. 1 kHz rather than
+# the 2 kHz physics rate because that is the real Kinova arm's low-level servo period
+# (jimmyyhwu/tidybot, robot/kinova.py: MoveJController.CYCLIC_STEP_SIZE)
+CONTROL_SCHEDULE_TIMESTEP = 0.001  # (in seconds)
+
 # Set macros needed for MuJoCo rendering
 _SYSTEM = platform.system()
 if _SYSTEM == "Windows":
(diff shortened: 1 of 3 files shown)


yixuanhuang98 (approved):
LGTM

PR #161, opened 2026-08-16
Add count-parameterized dynamic3D task family envs

yichao-liang:
## Why

Every kinder family except dynamic3D takes its object count as a constructor argument:

```python
Obstruction2D(num_obstructions=3)
DynScoopPour2D(num_small_circles=20)
Table3D(num_cubes=4)
```

The dynamic3D envs instead take a `task_config_path` naming one JSON file, and each file bakes in a count. A caller who wants to sweep an object count over one family therefore has to know how `tasks/` is laid out and assemble the filenames itself:

```python
TidyBot3DEnv(num_objects=4, task_config_path=".../tasks/ConstrainedCupboard3D/ConstrainedCupboard3D-o4.json")
```

Generalization experiments, where a single policy must run across instance sizes, need exactly that sweep, so today each consumer reimplements the same filename logic.

## What

`kinder.envs.dynamic3d.task_families` gives the seven families that register more than one count the usual constructor form:

```python
from kinder.envs.dynamic3d.task_families import ConstrainedCupboard3DEnv, SweepSimple3DEnv

ConstrainedCupboard3DEnv(num_objects=4)          # 1..6
SweepSimple3DEnv(num_objects=50)                 # 1, 5, 10, 50
```

Covered: ConstrainedCupboard3D, Dynamo3D, Shelf3D, Tossing3D, ScoopPour3D, SweepSimple3D, SortClutteredBlocks3D.

**Instructions.** Some families' variants differ by instruction as well as by count. Those carry a `default_instruction` that holds the goal fixed while the count varies, which is the property a count sweep depends on:

```python
SweepSimple3DEnv.available_instructions(50)
# ['sweep_the_blocks_to_the_left_side_of_the_kitchen_island',
#  'sweep_the_blocks_to_the_right_side_of_the_kitchen_counter', ...]
SweepSimple3DEnv(num_objects=50, instruction="sweep_the_blocks_to_the_right_side_of_the_kitchen_island")
```

The defaults are the only instruction registered at every count of their family (sweep-left-on-island; sort-into-bins) — the alternatives are single-count, so they offer no count axis.

**Robot base class.** Each family class mixes in ahead of the robot env class its task JSONs declare. `test_every_family_subclasses_the_env_its_tasks_declare` asserts that pairing against the JSONs, so a family that switched robots (all seven are `tidybot` today; `FrankaPickPlace3D` is `fr3`) cannot silently be built on the wrong base and fail only at reset.

**Shared robot map.** `ROBOT_ENV_CLASSES` moves out of `_register_dynamic3d`'s body into `kinder.envs.dynamic3d`, so the gym registration and these classes read one definition instead of two copies. It sits in the package `__init__` rather than beside the envs so registration can read it without importing mujoco, which is optional; `_register_dynamic3d` imports it locally for the same reason.

## Deliberately not covered

- **BalanceBeam3D, SweepIntoDrawer3D** — one registered variant each, so there is no count axis.
- **Rearrange3D** — its `o2` variant swaps in a different object (a can beside the boxed drink) and rewords the goal, rather than adding more of one object; the variants are two tasks, not one axis.

For all three, `gymnasium.make("kinder/<Family>-<variant>-v0")` is already the whole story.

Two labelling quirks the classes document rather than hide: `Dynamo3D-o12` places eleven chairs drawn from three chair models, and `SortClutteredBlocks3D-o12` counts three cuboids plus nine cubes. The `o` label names the variant; it is not a promise about object counts.

## Testing

`tests/envs/dynamic3d/test_task_families.py` — 10 tests covering: every declared count resolves to an existing task file; each family's base matches its JSONs' robot; default instructions span every count of their family; instruction discovery; unregistered counts / instructions and an explicit `task_config_path` all fail at construction with an actionable message; and construct-and-reset at two counts per family asserting the object count.

- New tests: 10 passed
- `tests/envs/dynamic3d/`: 279 passed, 4 skipped
- black, isort, mypy clean; pylint 10.00/10 on the changed files
- Run with `DISABLE_AUTO_DYNAMIC3D_SCENES_DOWNLOAD=1` and `scene_bg=False`, as CI does

Verified on macOS arm64 (Python 3.11); Linux is unverified locally, so CI is the real check there.


merlerm:
seems good to me but I'd rather give @tomsilver final approval here since I'm less familiar with the kinder code. In general I think we were considering moving away from the jsons all together at some point but that's probably a lot of work

src/kinder/envs/dynamic3d/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""Dynamic (MuJoCo) 3D environments.
```
tomsilver: Minor: as of recently, these dynamic3d environments are no longer just mujoco -- there are also limb manipulation environments using pybullet for dynamics. Just a comment update.

src/kinder/envs/dynamic3d/task_families.py
```diff
@@ -0,0 +1,204 @@
+"""Count-parameterized access to the registered dynamic3D task families.
+
+Every dynamic3D task ships as its own JSON under ``tasks/<Family>/``, named
```
tomsilver: This is also not necessarily true for all dynamic3D environments. Small comment update.

src/kinder/__init__.py
```diff
@@ -440,6 +440,12 @@ def _register_kinematic3d_v2() -> None:
 
 def _register_dynamic3d() -> None:
     # Tasks with different scenes and object counts
+    # Imported here rather than at module scope: this function only runs when the
```
tomsilver: I'm confused about the need for these changes and also the need for `ROBOT_ENV_CLASSES` to move. Isn't mujoco always installed?

tomsilver (approved):
Thanks a lot! One actual question and then a few minor comments

src/kinder/envs/dynamic3d/__init__.py
yichao-liang: Good catch. This file is gone entirely now.

src/kinder/envs/dynamic3d/task_families.py
yichao-liang: Fixed!

src/kinder/__init__.py
yichao-liang: You're right, and my comment was wrong: mujoco~=3.3.0 is a hard dependency, so there's no optional-backend concern to justify it.

yichao-liang:
Rebased onto main and addressed all three. The net diff is now just the two new files (task_families.py and its tests), nothing pre-existing was modified.

PR #166, opened 2026-08-19
kinder: attach Tossing3D's goal region to the bin instead of the ground

joshnroy:
# Human Written

This pr does two things:

1. Randomize the bin's spawn point (see the videos below)
2. Anchros (in json) the blocks goal region to the bin)

The amount of code is not too high, so in depth is not needed

image

https://github.com/user-attachments/assets/a8f4a382-ae2e-4682-8058-718b0df9b844

https://github.com/user-attachments/assets/dd47959a-dda2-4d2a-afa8-0e38da8f68eb

# AI

## Summary

Stacked on #162 -- built against its tip (`e3c5c98`) since that's the commit with the
tightened `blocks_goal_region` this fix's coordinates are derived from; `270fdb6` (the
tightening) isn't on kindergarden's own `main` yet.

`blocks_goal_region` had `"target": "ground"` with absolute world-frame coordinates --
zero structural link to `bin_0`, even though the bin itself is already a real, movable
object (freejoint, already round-trips through `set_state`). This is the same failure
class as `1183de7`: move the bin (e.g. randomize its placement at reset) and the scored
box stays behind. Fixes it by reparenting the region onto the bin's own body so it
tracks live via the kinematic tree, no new bookkeeping needed.

Also extends `MujocoObject._create_regions()` to accept 6-value ranges (previously only
`MujocoGround`'s did) -- a region parented on an object with just a 4-value range got a
hardcoded +/-1cm z-band around the object's local origin, which is far too thin for a
cube resting inside the bin (bin wall alone is 2cm).

Also widens `bin_init_region` from a degenerate single point (`(2.0, 0.0)`) to a real
range (`[1.85,-0.20]`-`[2.15,0.20]`), so the bin's placement can actually vary across
seeds -- it already goes through the same `_initialize_object_poses` machinery
`cube_0` does, so no new randomization code was needed either.

## Test plan

TDD: wrote `test_the_goal_regions_live_bbox_moves_with_the_bins_own_position` first
(move the bin 0.30m via snapshot/restore, assert the goal region bbox moves with it),
confirmed it failed on unmodified code (bin moved, box didn't), then implemented the
fix and confirmed it passes.

`bin_init_region`'s widened range re-checked against the tightened scoring window with
a 10-seed oracle-episode probe -- all 10 still solve, since the toss skill targets the
live goal region rather than a fixed offset from the bin.

Full `tests/envs/dynamic3d/` suite: 397 passed, 4 skipped.

## Followup

None from this side. A companion `kinder-baselines` PR swaps the goal-region check off
the ground fixture onto the bin object to match. The downstream `hitl-pmp` repo has its
own matching change ready (not yet pushed/pinned).

### DIFF
--- a/src/kinder/envs/dynamic3d/envs.py
+++ b/src/kinder/envs/dynamic3d/envs.py
@@ -1212,6 +1212,10 @@ class ObjectCentricRobotEnv(ObjectCentricDynamic3DRobotEnv[TidyBot3DConfig]):
         """Set the camera to use for rendering."""
         self._render_camera_name = camera_name
 
+    def get_object(self, name: str) -> MujocoObject:
+        """Look up a live scene object by name."""
+        return self._objects_dict[name]
+
     @abc.abstractmethod
     def _get_object_centric_robot_data(self) -> dict[Object, dict[str, float]]:
         """Get object-centric data for the robot.
--- a/src/kinder/envs/dynamic3d/objects/base.py
+++ b/src/kinder/envs/dynamic3d/objects/base.py
@@ -321,33 +321,46 @@ class MujocoObject:
     def _create_regions(self) -> None:
         """Create Region objects with site elements for each region.
 
-        Each region's 2D ranges [x_start, y_start, x_end, y_end] are converted to 3D
-        bounding boxes [x_min, y_min, z_min, x_max, y_max, z_max] where the z dimension
-        spans a small height above the object surface.
+        Each region's ranges are converted to 3D bounding boxes
+        [x_min, y_min, z_min, x_max, y_max, z_max]. A 4-value range
+        [x_start, y_start, x_end, y_end] gets a small, fixed z band around the object's
+        own local origin (the historical behaviour, still right for a region that only
+        needs to test "on top of this object's surface"). A 6-value range
+        [x_start, y_start, z_start, x_end, y_end, z_end] carries its own z bounds instead
+        -- needed for a region that must span real height in the parent's local frame,
+        e.g. a goal region attached to a container object rather than to the ground.
         """
         assert self.regions is not None, "Regions must be defined"
         assert (
             self.xml_element is not None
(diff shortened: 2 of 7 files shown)


src/kinder/envs/dynamic3d/envs.py
```diff
@@ -1212,6 +1212,10 @@ def set_render_camera(self, camera_name: str | None) -> None:
         """Set the camera to use for rendering."""
         self._render_camera_name = camera_name
 
+    def get_object(self, name: str) -> MujocoObject:
```
joshnroy: This is used by https://github.com/Princeton-Robot-Planning-and-Learning/kinder-baselines/pull/136 to avoid having to access internal attributes directly

src/kinder/envs/dynamic3d/objects/base.py
```diff
@@ -321,33 +321,43 @@ def __init__(
...
         placement_threshold = 0.01  # 1cm tolerance for placement
-        # Note: we are currently hard-coding the z range for the bounding boxes
-        # This could potentially be made configurable in the future.
 
         for region_name, region_config in self.regions.items():
             region_list: list[Region] = []
 
             for region_idx, region_range in enumerate(region_config["ranges"]):
-                x_start, y_start, x_end, y_end = region_range
+                if len(region_range) == 4:
```
yixuanhuang98: magic number : )

src/kinder/envs/dynamic3d/objects/base.py
```diff
@@ -321,33 +321,43 @@ def __init__(
...
+                    x_start, y_start, x_end, y_end = region_range
+                    z_start, z_end = 0.0, 0.0
+                elif len(region_range) == 6:
```
yixuanhuang98: ditto

yixuanhuang98 (approved):
Left some comments!

PR #168, opened 2026-08-20
Require the cubes to be on the beam for BalanceBeam3D to be solved

yichao-liang:
`BalanceBeam3D-o3`'s goal was `[["balanced", "seesaw_1", 5.0]]` -- the beam within 5 degrees of level, and nothing else. An empty beam already is, so the goal held at reset and **the task was solved by doing nothing**. kinder's own generated docs record it:

> **Random Action Stats**: Total Reward: -1.00, Success: **Yes**, Steps: **1**

The variant description says what was meant: *"Place three cubes onto a seesaw without tipping it past the balance threshold."* Only the second half was ever checked.

Reproduced through the wrapper robocode evaluates with, on every seed:

```
seed 0 zero-action step -> terminated=True  reward=-1.0
seed 1 zero-action step -> terminated=True  reward=-1.0
seed 2 zero-action step -> terminated=True  reward=-1.0
```

Tossing3D, SweepIntoDrawer3D and Obstruction3D all give 0.00 under a random policy, so this is specific to BalanceBeam3D rather than something systemic.

## Stating the task

No new predicate -- the goal DSL already says this. A `beam_goal_region` attached to the seesaw, one `on` atom per cube, conjoined with the balance check:

```json
"beam_goal_region": {
    "target": "seesaw_1",
    "rgba": [0.2, 0.6, 0.2, 0.4]
},
```
```json
"goal_state": [
    "and",
    ["on", "small_block_1", "beam_goal_region"],
    ["on", "small_block_2", "beam_goal_region"],
    ["on", "large_block", "beam_goal_region"],
    ["balanced", "seesaw_1", 5.0]
]
```

Same shape as #166's Tossing3D goal region on `bin_0`. It also inherits region rendering, so the target is visible in the scene instead of implicit.

## The region derives its extents from the seesaw

Note what the region does **not** say: any ranges. A seesaw fills those in from the beam it actually builds, via `GeneratedSeesaw.beam_surface_range()`, so a task never restates `beam_length` or `beam_width` and the region cannot go stale when that geometry changes.

Two choices in it worth flagging:

**The floor sits at the pivot, not at the beam's top surface.** The beam tilts. A cube near the end rides up to `beam_length/2 * sin(5 deg)` = 1.6cm below the level surface at the tolerance this task calls balanced -- more than the 1cm of slack `_create_regions` adds. The pivot is still well clear of anything lying on the ground beside or beneath the beam, and there is a test for that case.

**The ceiling is generous.** Gravity decides what is resting on the beam, so the upper bound only has to exclude something held far above it.

## `is_object_on_beam` was measuring against the world axes

Separately: the helper compared world x against half the beam *length* and world y against half its *width*. This task yaws the seesaw 90 degrees, so the two limits were swapped -- 0.18m of beam judged against a 0.035m half-width, and a cube resting mid-beam read as off it:

```
along=-0.126 -> fixed=True  old(world-axis)=False
along=+0.126 -> fixed=True  old(world-axis)=False
```

It now rotates the offset into the seesaw's frame first. Yaw only, deliberately: folding in the hinge pitch would shrink the footprint a cube has to sit within simply because the beam is not level, which is backwards for a task about loading a beam that tilts.

The helper had **no caller in the repo**. The existing o4 test does reach it through `_assert_block_on_seesaw`, but pins the seesaw to yaw 0, which is why this survived.

## Verification

Five tests, **each checked to fail without the change it covers**:

| test | fails without |
|---|---|
| doing nothing does not solve the shipped task | the goal change |
| cubes placed to balance the beam **do** solve it | goal change + orientation fix |
| the region's extents come from the beam | the derived region |
| a cube beside, past, or under the beam is not in the region | the derived region |
| the beam's extents follow the beam, not world x | the orientation fix |

The second matters as much as the first: tightening a goal is only correct if the task stays achievable, so it places the cubes, settles the physics, and asserts the goal holds.

`400 passed, 4 skipped` across `tests/envs/dynamic3d`; pylint and mypy clean.

## Not included

`docs/envs/BalanceBeam3D.md` and the variant page are generated by `scripts/docs/generate_env_docs.py`, so they are untouched here. Regenerating them will drop the "Success: Yes, Steps: 1" line quoted above.

### DIFF
--- a/src/kinder/envs/dynamic3d/objects/generated_objects.py
+++ b/src/kinder/envs/dynamic3d/objects/generated_objects.py
@@ -8,6 +8,7 @@ from pathlib import Path
 import numpy as np
 from numpy.typing import NDArray
 from relational_structs import Object
+from scipy.spatial.transform import Rotation
 
 from kinder.envs.dynamic3d.mujoco_utils import MujocoEnv
 from kinder.envs.dynamic3d.object_types import MujocoMovableObjectType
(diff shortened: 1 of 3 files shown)


yixuanhuang98 (approved):
LGTM

